PR #6, opened 2016-06-19
Fix path duplication by using root path on urls

djalmaaraujo:
When you go deep into more directories, the path gets duplicated because the url's doesn't start with /.


rauchg:
Great catch @djalmaaraujo

djalmaaraujo:
@rauchg :) 
I was in debt with by 1 missing PR, hehe.

PR #18, opened 2016-09-29
Always load Font Awesome CSS over HTTPS

mathiasbynens:
There is no reason not to load the Font Awesome CSS file over HTTPS in all cases.

Protocol-relative URLs are an anti-pattern. https://github.com/konklone/cdns-to-https#conclusion-cdns-should-redirect-to-https


rauchg:
Good call. We also will be inlining the SVGs directly soon

PR #20, opened 2016-09-29
Add `rel=noopener` with fallback

mathiasbynens:
This ensures that github.com can not mess with the `list`-powered HTML document in any way.

See https://mathiasbynens.github.io/rel-noopener/ for more information.

### DIFF
--- a/views/index.hbs
+++ b/views/index.hbs
@@ -31,7 +31,7 @@
     </main>
 
     <aside>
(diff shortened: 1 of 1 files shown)


mathiasbynens:
An alternate solution (that I’d personally prefer) is to just drop `target=_blank` instead.

rauchg:
On 🔥

PR #21, opened 2016-09-29
README: Fix uninstall step in `contribute` section

jfmengels:
The text says to uninstall `now-serve`, but the command uninstalls `list`.

Thanks for the tool, it works really well :)


leo:
Hihi, whoopsie! ❤️

PR #28, opened 2016-10-10
Fix route links

pablopunk:
screen shot 2016-10-10 at 5 25 22 pm

If I click a link in the route (see image) it will take me to http://link. So for example, if I click on `repos` I would be redirected to [http://repos](http://repos).

This simple change fixed it, but I'm not sure if this affects some other code.


leo:
Thanks! 😊

PR #34, opened 2016-10-24
Send custom 404 page if it exists

taravancil:
This checks if `404.html` exists in the current directory and on 404s sends that instead of the default `Not Found` response.

### DIFF
--- a/bin/list
+++ b/bin/list
@@ -172,8 +172,17 @@ const handler = async (req, res) => {
 
   related = decodeURIComponent(path.format(related))
 
+  let notFoundResponse
+
+  try {
+    const custom404Path = path.join(current, '/404.html')
+    notFoundResponse = fs.readFileSync(custom404Path, 'utf-8')
+  } catch (err) {
+    notFoundResponse = 'Not Found'
+  }
+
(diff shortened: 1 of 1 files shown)


bin/list
```diff
@@ -172,16 +172,25 @@ const handler = async (req, res) => {
 
   related = decodeURIComponent(path.format(related))
 
+  let notFoundResponse
+
+  try {
+    const custom404Path = path.join(current, '/404.html')
+    notFoundResponse = fs.readFileSync(custom404Path, 'utf-8')
```
leo: Can you make this use `await`? Like [here](https://github.com/zeit/list/blob/master/bin/list#L211).

bin/list
```diff
@@ -172,16 +172,25 @@ const handler = async (req, res) => {
...
+  } catch (err) {
+    notFoundResponse = 'Not Found'
```
leo: I also recommend making this `catch (err) {}` and set `'Not Found'` as the default value of `notFoundResponse` instead.

taravancil:
Cool! I just pushed those changes. Lmk if you need any others 😊

leo:
Thanks a lot - that was all! 😊

PR #37, opened 2016-10-26
Add --cors option to allow access from any origin

slorber:
Issue: https://github.com/zeit/list/issues/36

Here's a support for cors headers.

I did take the same setting in source of http-server for the same --cors option. Maybe it can be tweaked in the future but it's a good start.

Note that maybe the --cors option short version declaration (with args) should be reworked (see https://github.com/leo/args/issues/37)

### DIFF
--- a/bin/serve
+++ b/bin/serve
@@ -25,6 +25,7 @@ args
   .option('unzipped', 'Disable GZIP compression')
   .option('ignore', 'Files and directories to ignore', '')
   .option('auth', 'Serve behind basic auth')
+  .option(['o','cors'], 'Setup * CORS headers to allow requests from any origin',false)
 
 const flags = args.parse(process.argv)
 const directory = args.sub[0]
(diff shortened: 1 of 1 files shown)


leo:
@slorber Please rebase 😊

slorber:
done

leo:
Thanks!

slorber:
@leo by chance is this usable with now --static command?

leo:
@slorber Link mentioned in [here](https://zeit.co/blog/now-static), you need to add a custom scripts property to your `package.json` file and run `list` if you want to use it.

slorber:
Yeah I already do that (with now-build and now-start), but reverted back to building locally for some reasons (takes very long time to build on zeit, optipng not working when building on zeit, also not very precise/documented control over what is uploaded regarging .gitignore/.npmignore/zeit.files... we also got some 502 errors on such deployment but according to your team it may be hardware related)

Currently I'm just using this package.json that I put inside /build folder of my app

```
{
  "name": "stample-front",
  "description" : "For Zeit.co / Now static frontend deployment of Stample",
  "scripts": {
    "start": "list --cors --cache 1"
  },
  "dependencies": {
    "list-seb": "0.8.3"
  }
}
```

I was just wondering if I could be able to replace it with a command line, like `now --static --cors --cache 1`

PR #41, opened 2016-10-26
Better etag/304 support

slorber:
Here is basic support for etas 304

https://github.com/zeit/list/issues/40

### DIFF
--- a/bin/list
+++ b/bin/list
@@ -161,6 +161,34 @@ const renderDirectory = async dir => {
   return render(details)
 }
 
+
+
+// To enhance, one can take inspiration from https://github.com/jfhbrook/node-ecstatic/blob/fe17a5623c18319f82101b6549e9e2e526e03d1f/lib/ecstatic.js#L284
+const isNotModified = (req, etag, fileStats) => {
+  try {
+    // Etag comparison
+    const requestEtag = req.headers['if-none-match'];
+    if (requestEtag && (requestEtag === etag)) {
+      return true;
+    }
+    // Last-modified comparison
+    // TODO: Original code, left untouched, but for me it's not working at all
+    // The algo seems wrong to be and we don't even set a Last-modified header in response
+    let requestDate = req.headers['if-modified-since']
+    if (requestDate) {
+      requestDate = new Date(requestDate)
+      if (requestDate.getTime() === fileStats.mtime.getTime()) {
+        return true;
+      }
+    }
+  } 
+  catch (e) {
+    return false; // failsafe in case of failure
+  }
+  return false;
+}
+
+
 const handler = async (req, res) => {
   const {pathname} = parse(req.url)
   let related = path.parse(path.join(current, pathname))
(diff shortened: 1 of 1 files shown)


bin/list
```diff
@@ -234,42 +262,33 @@ const handler = async (req, res) => {
...
-
   for (const header in defaultHeaders) {
     if (!{}.hasOwnProperty.call(defaultHeaders, header)) {
       continue
     }
-
     res.setHeader(header, defaultHeaders[header])
   }
 
+  if (isNotModified(req,etag,stats)) {
```
fearphage: Spacing is inconsistent with the rest of the file.

bin/list
```diff
@@ -234,42 +262,33 @@ const handler = async (req, res) => {
...
-  let statusCode = 200
-
-  if (requestDate) {
-    requestDate = new Date(requestDate)
-
-    if (requestDate.getTime() === stats.mtime.getTime()) {
-      statusCode = 304
-    }
-  }
+  const etag = '"' + stats.dev + '-' + stats.ino + '-' + stats.mtime.getTime() + '"'
```
fearphage: Use a template string.

bin/list
```diff
@@ -161,6 +161,34 @@ const renderDirectory = async dir => {
   return render(details)
 }
 
+
+
+// To enhance, one can take inspiration from https://github.com/jfhbrook/node-ecstatic/blob/fe17a5623c18319f82101b6549e9e2e526e03d1f/lib/ecstatic.js#L284
+const isNotModified = (req, etag, fileStats) => {
+  try {
```
fearphage: What do you expect to `throw` here? I find it best to put as little as possible in `try` blocks... like usually the one line that can throw. This makes it obvious to other developers where the problems lie.

bin/list
```diff
@@ -161,6 +161,34 @@ const renderDirectory = async dir => {
...
+    let requestDate = req.headers['if-modified-since']
+    if (requestDate) {
+      requestDate = new Date(requestDate)
+      if (requestDate.getTime() === fileStats.mtime.getTime()) {
+        return true;
+      }
+    }
+  } 
+  catch (e) {
+    return false; // failsafe in case of failure
```
fearphage: Can't you just do nothing here and let the next line handle it?

bin/list
```diff
@@ -161,6 +161,34 @@ const renderDirectory = async dir => {
...
 }
 
+
+
+// To enhance, one can take inspiration from https://github.com/jfhbrook/node-ecstatic/blob/fe17a5623c18319f82101b6549e9e2e526e03d1f/lib/ecstatic.js#L284
+const isNotModified = (req, etag, fileStats) => {
+  try {
+    // Etag comparison
+    const requestEtag = req.headers['if-none-match'];
+    if (requestEtag && (requestEtag === etag)) {
```
fearphage: There is no need to check whether the `requestEtag` is defined, because `etag` is never `undefined` nor `null`. Thus `undefined` is guaranteed to not equal whatever `etag` is.

bin/list
```diff
@@ -161,6 +161,34 @@ const renderDirectory = async dir => {
...
+      return true;
+    }
+    // Last-modified comparison
+    // TODO: Original code, left untouched, but for me it's not working at all
+    // The algo seems wrong to be and we don't even set a Last-modified header in response
+    let requestDate = req.headers['if-modified-since']
```
fearphage: Was the code working before your changes? If it wasn't and still isn't working, what's the value of keeping it around (based on your comment)?

bin/list
slorber: @fearphage I don't know what can throw now, or what can throw in the future (if someone modifies the code), but I'm sure that failsafe behavior (caching disabled) is probably better than total unability to serve files

bin/list
slorber: that's right

bin/list
slorber: @fearphage I'm not from Zeit either so I won't refactor their own code and remove potential code I may not understand totally without a feedback from them

If it was my own code I would delete that immediately and will delete it as soon as someone from Zeit confirm it's useless

bin/list
slorber: I can too if someone from Zeit finally want this, but it was like that before

leo:
Thanks for your effort, guys! But we removed the whole transpilation step from the development workflow and the package content therefore changed completely.

Also I still don't understand what the advantage of this change is. I thought we already support browser caching perfectly?

slorber:
@leo I don't know what you mean by "supporting browser caching perfectly" but at least for me 304/etags did not work like intended.

I would be happy to backport my changes on the new code, or at least show you what does not work

leo:
@slorber Ahh, I see. Just read your issue. Feel free to open a new PR with all the changes necessary for making it work properly (no need to inherit/keep code we wrote).

PR #43, opened 2016-11-05
Serve files with no extension

### LINKED ISSUE #25: Returns 404 errors for files without extensions
I have a file I'm trying to deploy that doesn't have an extension. eg. /framer/version

The code doesn't see an extension, so it tries to find /framer/version/index.html, which doesn't exist.

https://github.com/zeit/list/blob/master/bin/list#L187

I was quite surprised by this behaviour.

taravancil:
Fixes #25

### DIFF
--- a/bin/list
+++ b/bin/list
@@ -86,12 +86,13 @@ const renderDirectory = async dir => {
     const index = files.indexOf(file)
     const details = path.parse(filePath)
 
-    details.ext = details.ext.split('.')[1]
     details.relative = path.join(subPath, details.base)
 
     if (await isDir(filePath)) {
       details.base += '/'
     } else {
+      details.ext = details.ext.split('.')[1] || 'txt'
+
       let fileStats
 
       try {
--- a/bin/list
+++ b/bin/list
@@ -185,12 +185,8 @@ const handler = async (req, res) => {
   }
 
   // Check if file or directory path
-  if (path.parse(related).ext === '') {
+  if (await isDir(related)) {
     let indexPath = path.join(related, '/index.html')
-
-    if (!await isDir(related) && flags.single === undefined) {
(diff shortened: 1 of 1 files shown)


TehShrike:
Sweeeet.  I am very excited for this to get merged and deployed!

bin/list
```diff
@@ -86,12 +86,13 @@ const renderDirectory = async dir => {
     const index = files.indexOf(file)
     const details = path.parse(filePath)
 
-    details.ext = details.ext.split('.')[1]
     details.relative = path.join(subPath, details.base)
 
     if (await isDir(filePath)) {
       details.base += '/'
     } else {
+      details.ext = details.ext.split('.')[1] || 'txt'
```
taravancil: For files with no extension, `details.ext.split('.')[1]` is `undefined`. `details.ext` is used [here](https://github.com/zeit/list/blob/master/views/index.hbs#L26) to add a class that determines which icon is used, and if it's `undefined`, [the class that renders the directory icon](https://github.com/zeit/list/blob/master/assets/styles.scss#L116) will apply.

Using `anyrandomstring` instead of `txt` would yield the same results, but I chose `txt` in order to be consistent with how`.txt` files are handled.

wolverian: Ah, thanks for the explanation! I asked because it was confusing. 👍

leo:
Thanks a lot! 😊

burtontanner:
I still cannot serve files without extensions. Im using version 11.0.2. Any Ideas?

discordianfish:
Same, I feel like there might be a regression?

discordianfish:
I think this broke it: https://github.com/zeit/serve-handler/commit/2b5d6a090bcefbce9cdcefe0d16e4afe6fed7d35

Issue #26, opened 2016-10-04
Pass custom items to `ignoredFiles` array


joshgillies:
I've noticed that currently `ignoredFiles` is a hard coded array, and wonder whether there's much value to offering a method to customise the items in this list? The addition of an `--ignore` flag would suffice.

joshgillies:
An interesting idea just crossed my mind, what if `list` parsed `.gitignore` or `.npmignore` to base it's ignored files/folder off.

leo:
Fixed in https://github.com/zeit/list/pull/44

PR #44, opened 2016-11-05
add option to ignore custom files and directories

taravancil:
This implements an option to customize which files and directories to ignore, as suggested in #26.

I opted to only add an `--ignore` flag instead of parsing `.gitignore` or `.npmignore`. Given how I  personally use `list`, I'd find it inconvenient if everything in my `.gitignore` was not served. I'm happy to implement it though if y'all think that's appropriate!


leo:
Looks good to me! Thanks 😊

Issue #30, opened 2016-10-12
Basic password protection


guidobouman:
When using list, you might want to supply a password to protect your directory or static site.

This is especially important when using [now-serve](https://github.com/zeit/now-serve) to serve a staging environment that should not be reachable by every person on the internet.

guidobouman:
@leo I'm working on a PR, should this be in the form of a [micro](https://github.com/zeit/micro)-plugin or can I use [http-auth](https://www.npmjs.com/package/http-auth)?

rauchg:
This is a cool idea. Maybe with envs? `HTTP_BASIC_USER` and `HTTP_BASIC_PASSWORD` ?

guidobouman:
Yeah, I thought the same. Or with a `HTTP_BASIC_AUTH` string: `username:password` or `username:password_hash`

taravancil:
@guidobouman are you still working on this? I was thinking about taking a stab at it, but I don't want to jump in if you're still planning to do it 😄

guidobouman:
Have a go! I can assist if you're running into any issues.

taravancil:
Cool, thanks!

timkendall:
Implemented this as just more CLI flags. Let me know what you guys think.

leo:
https://github.com/zeit/list/pull/49

paulintrognon:
Hi, since the rewrite I can't see this option anywhere. Is it still possible to add a basic protection?

flawyte:
@paulintrognon Yes, but it seems like the only way now is to do it by yourself. See how [here](https://github.com/flawyte/now-basic-auth).

PR #49, opened 2016-11-30
Add HTTP Basic Auth support

timkendall:
#30

### DIFF
--- a/bin/list
+++ b/bin/list
@@ -24,8 +24,6 @@ args
   .option('unzipped', 'Disable GZIP compression')
   .option('ignore', 'Files and directories to ignore', '')
   .option('auth', 'Serve behind basic auth')
-  .option('username', 'Username for basic auth', 'example')
-  .option('key', 'Password for basic auth', '1234')
 
 const flags = args.parse(process.argv)
 const directory = args.sub[0]
@@ -176,7 +174,7 @@ const handler = async (req, res) => {
   if (flags.auth) {
     const credentials = auth(req)
 
-    if (!credentials || credentials.name !== flags.username || credentials.pass !== flags.key) {
+    if (!credentials || credentials.name !== process.env.LIST_USER || credentials.pass !== process.env.LIST_PASSWORD) {
       res.statusCode = 401
       res.setHeader('WWW-Authenticate', 'Basic realm="User Visible Realm"')
       return send(res, 401, 'Access Denied')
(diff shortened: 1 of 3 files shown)


leo:
Looks great! But please don't increment the version tag in `package.json` for us. Depending on what will follow, it might also be a major release... 😊

timkendall:
Gotcha. Was wondering about that. I'll revert it when I get back to my computer later today. Great little library thanks for sharing!

aesopwolf:
I'd be interested in seeing a slightly different API. One that pulls user/pass from environment variables automatically (perhaps `process.env.LIST_BASIC_AUTH_USER` and `process.env.LIST_BASIC_AUTH_PASS` by default). This would require only a single command line arg of `--auth`

You can then optionally supply a user/pass directly to the auth command in case you want to override looking up the environment variables. So `--auth username:password`

timkendall:
Hmm yeah this could be good. @leo any thoughts?

leo:
👌 

But let's only do the env vars, so that we can prevent people from hardcoding stuff. On now, env variables can be filled with secrets defined on the platform.

`LIST_USER`
`LIST_PASSWORD`

rauchg:
Not possible with basic auth, but that would be dreamy for many many scenarios :)

bin/list
```diff
@@ -168,6 +170,17 @@ const renderDirectory = async dir => {
 }
 
 const handler = async (req, res) => {
+  
+  if (flags.auth) {
+    const credentials = auth(req)
+
```
rauchg: It's missing error reporting if the env vars are not set

timkendall:
Oops. Hows that?

leo:
Thanks!

electerious:
How can I use HTTP Basic Auth together with `now`? The following isn't working.

```
now -e SERVE_USER=abc -e SERVE_PASSWORD=abc
```

leo:
@electerious The flag is missing: https://github.com/zeit/serve#authentication

electerious:
True, but how can I pass the `--auth` to the `serve` module when deploying via `now`?

derhuerst:
@electerious It should be just like to did above.

electerious:
But `now --auth -e SERVE_USER=abc -e SERVE_PASSWORD=abc` passes the `--auth` flag to `now`, right? It won't be forwarded to `serve`.

leo:
Correct. You can't pass flags to serve via `now`. You need a custom package.json

derhuerst:
@electerious The `start` script in `package.json` would usually contain the `--auth` flag. User & password would be read from the env, provided by you via `now -e …`.

t829702:
is this feature documented how to use standalone?
https://github.com/vercel/serve#authentication

PR #52, opened 2016-12-08
Fix for the `--single` option

domderen:
Right now running list with `--single` option doesn't work correctly. I tries to load the file under a specified path, and when it doesn't find it, it crushes with an error:

```
Error: ENOENT: no such file or directory, open '/Users/dominik/Downloads/theory-463759986/whatever'
    at Error (native)
```

The above commit reverts the line that introduced the bug and fixes the behaviour for `--single` option.

### DIFF
--- a/bin/list
+++ b/bin/list
@@ -170,7 +170,7 @@ const renderDirectory = async dir => {
 }
 
 const handler = async (req, res) => {
-  
+
   if (flags.auth) {
     const credentials = auth(req)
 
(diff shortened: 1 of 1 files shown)


leo:
Files without extensions also exist 😊

domderen:
Updated the PR, hopefully this one is better ;)

leo:
Thanks a lot!

Issue #22, opened 2016-10-03
Add feature to serve page.html on requests to site.com/page


flybayer:
## Use case

I have a static site built with `create-react-app`. I'm using [react-snapshot](https://github.com/geelen/react-snapshot) which does zero-config SSR by generating `html` files for each page. The resulting build output is:

```
build/
   index.html
   history.html
   ...
```

Unfortunately when serving this directory with `list`, requests to `site.com/history` fail. In order to load `history.html`, I have to request `site.com/history.html`.

**It'd be great if a flag or default feature would be added to support that.** :smiley: 
### My current workaround

``` bash
mkdir build/history
mv build/history.html build/history/index.html
```

timneutkens:
Isn't this what `-s, --single` is for?

flybayer:
As far as can tell `--single` redirects all requests to `index.html`. So a request to `site.com/history` will be served `index.html`. I want `history.html` to be served in this case.

timneutkens:
You're right I was too quick 😉

leo:
Doesn't `site.com/history` serve `history/index.html`?

flybayer:
@leo, yes, that's my workaround. But what I want is `site.com/history` to serve `history.html` if it's there because that's what `react-snapshot` generates for me. My workaround is a custom script that moves `history.html` to `history/index.html` (and for every other page on the site) after my build.

joshgillies:
As [`react-snapshot`](https://github.com/geelen/react-snapshot) is geared toward deployments to https://surge.sh, shouldn't this be a feature request against `react-snapshot`? Unless, of course @leo sees value in making `list` behave in the above manner?

flybayer:
That's a possibility, but I'm sure someone else will come along and want this behavior for some other use case 🙂

flybayer:
For `list` to be the go-to solution for static file hosting, I think it needs to be flexible and have a great out of the box experience. Many other static file servers like `http-server` already support this by default. 

Can anyone think of a reason this feature shouldn't be added? The only thing I've thought of is we'd have to decide which should have the priority, `history.html` or `history/index.html`, in the case both are present.

joshgillies:
@flybayer if `http-server` already offers support for your use case, what brought you to `list`?

[...]

PR #58, opened 2016-12-14
add --clean option to allow clean urls

danreeves:
Inspired by https://surge.sh and issue #22.

Adds `--clean` option which:
 * redirects to remove `/index.html` and `.html` from urls
 * resolves an extension-less url path to a fs path with an extension, e.g. `[domain]/path` → `[dir]/path.html`

### DIFF
--- a/bin/serve
+++ b/bin/serve
@@ -26,6 +26,7 @@ args
   .option('ignore', 'Files and directories to ignore', '')
   .option('auth', 'Serve behind basic auth')
   .option(['o', 'cors'], 'Setup * CORS headers to allow requests from any origin', false)
+  .option('clean', 'Serve HTML without file extensions')
 
 const flags = args.parse(process.argv)
 const directory = args.sub[0]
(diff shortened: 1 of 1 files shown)


bin/serve
```diff
@@ -204,7 +205,32 @@ const handler = async (req, res) => {
...
+    if (pathname.indexOf('.html') > -1) {
+      res.setHeader('Location', pathname.replace(path.extname(pathname), ''))
+      return send(res, 303)
+    }
+
+    // Look for /path.html
+    let cleanUrl = path.format({
+      name: related,
+      ext: '.html',
+    })
```
leo: There's an empty line missing here!

bin/serve
```diff
@@ -204,7 +205,32 @@ const handler = async (req, res) => {
...
+    if (fs.existsSync(cleanUrl)) {
+      pathExists = true
+      related = cleanUrl
+    }
+  }
+
+  if (!pathExists && flags.single === undefined) {
```
leo: Just do: `(!pathExists && !flags.single)`

bin/serve
```diff
@@ -204,7 +205,32 @@ const handler = async (req, res) => {
...
 
-  if (!fs.existsSync(related) && flags.single === undefined) {
+  let pathExists = fs.existsSync(related)
+
+  if (flags.clean) {
+    // Redirect from /path/index.html to /path
+    if (pathname.indexOf('index.html') > -1) {
+      res.setHeader('Location', path.dirname(pathname))
+      return send(res, 303)
+    }
```
leo: Another empty line missing here!

bin/serve
```diff
@@ -26,6 +26,7 @@ args
   .option('ignore', 'Files and directories to ignore', '')
   .option('auth', 'Serve behind basic auth')
   .option(['o','cors'], 'Setup * CORS headers to allow requests from any origin',false)
+  .option('clean', 'Serves HTML files without file extensions')
```
leo: Please change to:

> Serve HTML without file extensions

danreeves:
Thanks for the feedback, I've made the requested updates and rebased on master.

leo:
I thought about this for some time and noticed that most static site generators do it like this: You basically create a directory called "test" and put a `index.html` file inside it. This will result in the following, extensionless path: `/test`.

So each time you access `/test`, the `index.html` file inside the folder gets loaded.

Please do it like that!

Also huge thanks for the efforts you pushed into making this PR.

danreeves:
That's a shame, thanks for considering it!

Issue #23, opened 2016-10-04
`index.htm` isn't served as directory index


joshgillies:

terkelg:
Is this the reason why index.html can't link to files relative to the current directory? Like in this example https://dist-myryxjqyxh.now.sh/1, where index.html can't find the stylesheets and JavaScript files?

leo:
Not willing to support this, sorry! It's a freaking old extension originating from the time in the DOS world where only 3 extension chars were allowed. Please use .html!

PR #61, opened 2016-12-17
Add index.htm as directory index

sa-mm:
This is meant to address #23. If 'index.htm' is found, then that will be loaded instead of rendering a list of files in the directory.

Not sure if this is the best solution.


leo:
Thanks for your effort, but we completely changed the content of the package (in turn, you'll have to open a new pull request). 😊

PR #68, opened 2017-01-12
Copy url to clipboard

pablopunk:
This is an idea I had, let me know what you think.

I love that `now` copies the address to my clipboard, and as I use `serve` a lot to quickly share a folder, I would love to have that feature here. Also, with the [random port feature](https://github.com/zeit/serve/commit/d177add755a0500c465fdfc87dcbde6d58a3cb01) in the last version, I find this way easier to work with (cause before I would be always visiting `:3000`)

serve


leo:
That's a really great idea, thanks!

PR #71, opened 2017-01-16
feat: serve specified directory

KnisterPeter:
Serves specified directory, defaults to cwd

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -18,6 +18,7 @@ args
   .option('ignore', 'Files and directories to ignore', '')
   .option('auth', 'Serve behind basic auth')
   .option(['o', 'cors'], 'Setup * CORS headers to allow requests from any origin', false)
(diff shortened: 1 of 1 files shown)


matheuss:
You already can do that by running `serve path/to/directory` 🤔

KnisterPeter:
I've tried that and it failed in combination with `--single`.

KnisterPeter:
Maybe I misunderstood single, but I thought it might be there for serving a single index.html and some static resources (like images or js or css)

jamo:
Shouldn't we rather fix the current method for specifying folder to work with `--single` - than add a new option for something that existed 🤔

jamo:
Seems like this https://github.com/zeit/serve/blob/master/lib/index.js#L23 is just too naive - afaik the path to directory should be last argument...

KnisterPeter:
So lets abandon this.

PR #76, opened 2017-01-18
Add default port: 3000

### LINKED ISSUE #75: Include Default Port
Upgrading from `2.0` to `2.4` breaks my simple use case unless I specify a port.

```json
{
  "start-2.0": "serve dist -s",
  "start-2.4": "serve dist -s -p 3000"
}
```

It also breaks whether or not the `-s` flag is present.

lukeed:
Closes #75

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,8 +10,10 @@ const compress = require('micro-compress')
 const listening = require('./listening')
 const serverHandler = require('./server')
 
+const port = process.env.PORT || 3000
+
 args
-  .option('port', 'Port to listen on', process.env.PORT)
+  .option('port', 'Port to listen on', port)
   .option('cache', 'How long static files should be cached in the browser (seconds)', 3600)
   .option('single', 'Serve single page apps with only one index.html')
   .option('unzipped', 'Disable GZIP compression')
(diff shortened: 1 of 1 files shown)


leo:
I removed this because I wanted the default port to be a random one.

lukeed:
Oh, okay. I was thrown off because it doesn't work at all without explicitly setting the flag or env var.

leo:
@lukeed What's the node version you're on it? Because if you don't pass a port to a HTTP server in node, it will normally just use a random open port.

lukeed:
I'm on `6.9.1` 🤔

leo:
@lukeed What's the error you're getting?

lukeed:
This happened to me on a `now` deployment as well, even when `serve` was the only dependency installed. So I don't think I'm alone here?

```
RangeError: "port" argument must be >= 0 and < 65536
    at assertPort (internal/net.js:17:11)
    at Server.listen (net.js:1388:5)
    at Object.<anonymous> (/Users/lukee/repos/personal/apps/colors-app/node_modules/serve/lib/index.js:47:8)
    at Module._compile (module.js:570:32)
    at Module._compile (/Users/lukee/repos/personal/apps/colors-app/node_modules/async-to-gen/register.js:37:23)
    at Module._extensions..js (module.js:579:10)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/lukee/repos/personal/apps/colors-app/node_modules/async-to-gen/register.js:40:5)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
```

leo:
@lukeed Fixed! Please get the latest update.

lukeed:
Cool, thanks!

PR #77, opened 2017-01-19
Display IP address instead of localhost

disquisition:
Display network IP address (rather than localhost) both on command line and in rendered view to make sharing links across machines more seamless.


leo:
🙏

PR #78, opened 2017-01-19
Fix asset dir check on Windows

disquisition:
The asset dir check always fails on Windows unless you normalize the delimiters. This bug resulted in the stylesheet 404ing when running `serve` from a Windows machine.


leo:
Thanks a lot!

PR #80, opened 2017-01-20
enable listen on process.env.PORT

lukeed:
Currently lists the `process.env.PORT` as the default port but it was never actually passed to the server.

Order of selection is now `flags.port > process.env.PORT > randomPort` .

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -11,8 +11,10 @@ const getPort = require('get-port')
 const listening = require('./listening')
 const serverHandler = require('./server')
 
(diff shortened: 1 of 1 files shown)


leo:
@lukeed It was the default value before, so it `args` sat it as the value of `flags.port`. Isn't that correct? At least that's how it should work.

lukeed:
Oh, correct 🤔 Not sure why I didn't see this happen the first time around -- probably a typo. It works as is.

PR #81, opened 2017-01-20
Prevent throw during `--single` mode

### LINKED ISSUE #73: --single breaks now.
It seems like it's throwing when the directory doesn't exist:

```js
if (await pathType.dir(related)) { // ...
```

This used to work, but now it's broken :(

I'm simply using `serve ./build --single` and a `react-app` created app.

lukeed:
The `path-type` module does not wrap its `dir()` functions with any kind of `catch` handler, so it requires that all strings must be pre-verified. 

During `--single` mode, any nested route (eg, `/blog/article2`) is eventually passed along as `/Users/.../dist/blog/article2`, which does not exist (assuming `dist` is the base directory).

And so, because non-existent paths are sent to `pathType.dir()`, a `Error: ENOENT` is thrown, which causes the 502 error that I & others were experiencing. 

Closes #73

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -43,6 +43,8 @@ module.exports = async (req, res, flags, current, ignoredFiles) => {
 
   related = decodeURIComponent(path.format(related))
 
+  const relatedExists = fs.existsSync(related)
+
   let notFoundResponse = 'Not Found'
 
   try {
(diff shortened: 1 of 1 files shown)


lib/server.js
```diff
@@ -43,19 +43,21 @@ module.exports = async (req, res, flags, current, ignoredFiles) => {
 
   related = decodeURIComponent(path.format(related))
 
+  const relatedExists = fs.existsSync(related)
```
leo: Can you do `await fs.exists()` please? Not sure if that will throw or not. Wrap it in `try` if it will... 😊

lukeed: Unfortunately `fs.exists` is deprecated. Should I use a module instead?

leo: @lukeed Ahhhh, you're right. Nope!

lukeed:
Thanks!

PR #83, opened 2017-01-23
Fix directory name render issue when running on Windows

disquisition:
Current logic surrounding generating the rendered directory name uses hardcoded directory separator; this resulted in the directory name missing from rendered output when running `serve` on Windows. This change updates the render code to instead use OS-specific separator.


leo:
Great idea, thanks!

PR #93, opened 2017-02-05
Correct pseudo-elements syntax

electerious:
Makes no difference as all browser support the old syntax for backward compatibility, but I prefer the correct syntax.

```
The double colon replaced the single-colon notation for pseudo-elements in CSS3. This was an attempt from W3C to distinguish between pseudo-classes and pseudo-elements.

The single-colon syntax was used for both pseudo-classes and pseudo-elements in CSS2 and CSS1.

For backward compatibility, the single-colon syntax is acceptable for CSS2 and CSS1 pseudo-elements.
```

### DIFF
--- a/assets/styles.css
+++ b/assets/styles.css
@@ -30,7 +30,7 @@ h1 a {
   border-bottom: 1px dashed transparent;
 }
 
-h1 a:after {
+h1 a::after {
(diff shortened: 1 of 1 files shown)


leo:
TIL. Thx!

PR #104, opened 2017-02-11
Remove the bullet characters

thorn0:
On Windows, they're not printed to the console and produce an unexpected beep sound instead.


leo:
Oh. Thanks! 😅

PR #106, opened 2017-02-11
Add title attributes to file links

disquisition:
This change adds `title` attributes to the file links such that when the file name is elided, the user can hover over the link to see the full file name.

add-link-titles

I've added the `title` property to the file objects rather than simply re-using the `base` property because I wanted to be able to add more verbose titles to the parent directory links.


leo:
WOW! That's really great. Thanks a lot!

PR #107, opened 2017-02-11
Remove ip package from render code

disquisition:
The IP address was removed from the view but not from the render code.


leo:
👌 🙏

PR #109, opened 2017-02-13
Fix a trailing slash with query strings

nulltask:
I've fixed unexpected normalizing trailing slash.

- before: `/foo?bar=baz` -> `/foo?bar=baz/`
- after: `/foo?bar=baz` -> `/foo/?bar=baz`

Please review.


leo:
Thanks! 🙏

nulltask:
Thanks for your merging!

PR #112, opened 2017-02-16
Add command example in doc to set env variables

Fenwil:
Sometimes I want to launch serve quickly with some user and password and there was no copy-paste example in the documentation and without having to write in a file the environment variables.


leo:
190bc0d996c27a9a342bcb54000e6beeb855a9fe

PR #113, opened 2017-02-16
Add a programmatic API (fixes #86)

### LINKED ISSUE #86: Programmatic API
For loading `serve` inside an app

dar5hak:
This is a prototype of `api.js` as discussed in #86. Here's what I think it should do:

- [x] Export a function that takes `options` and `directory` as arguments
- [x] Start a server based on those
- [x] Print the server info nicely, with colours and stuff (tested on Linux)
- [x] Take `stdin`, such as <kbd>^C</kbd> and behave accordingly
- [x] Return something to the caller (what?)
- [x] Write docs

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -15,7 +15,14 @@ const optionsToArgs = options => {
 
 module.exports = (options = {}, directory = process.cwd()) => {
   const scriptPath = path.join(__dirname, '..', 'bin', 'serve.js')
-  spawn('node',
+
+  const cli = spawn('node',
     [scriptPath, ...optionsToArgs(options), directory],
     {stdio: 'inherit'})
+
+  return {
+    stop() {
+      cli.kill()
+    }
+  }
 }
--- a/lib/api.js
+++ b/lib/api.js
@@ -8,11 +8,17 @@ const dargs = require('dargs')
 module.exports = (options = {}, directory = process.cwd()) => {
   const scriptPath = path.join(__dirname, '..', 'bin', 'serve.js')
   const aliases = {cors: 'o'}
+
   options._ = [directory]  // Let dargs handle the directory argument
 
-  const cli = spawn('node',
-    [scriptPath, ...dargs(options, {aliases})],
-    {stdio: 'inherit'})
+  const args = [
+    scriptPath,
+    ...dargs(options, {aliases})
+  ]
+
+  const cli = spawn('node', args, {
+    stdio: 'inherit'
+  })
 
   return {
     stop() {
--- a/lib/api.js
+++ b/lib/api.js
@@ -2,22 +2,16 @@
 const {spawn} = require('child_process')
 const path = require('path')
 
-// { port: 1337 } => ['--port', '1337']
-const optionsToArgs = options => {
-  const args = []
-  for (const prop in options) {
-    if (Object.prototype.hasOwnProperty.call(options, prop)) {
-      args.push(`--${prop}`, options[prop])
-    }
-  }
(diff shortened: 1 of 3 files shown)


dar5hak:
Do you think this list is comprehensive? I'm still trying to figure out 4. Any hints on 5?

leo:
@dar5hak Please return an object with a `.stop()` method. This will solve both 4 and 5 as it allows people to stop the server programmatically. 🙏

lib/api.js
```diff
@@ -0,0 +1,21 @@
+// Native
+const {spawn} = require('child_process')
+const path = require('path')
+
+// { port: 1337 } => ['--port', '1337']
+const optionsToArgs = options => {
```
tunnckoCore: can use `dargs` directly

dar5hak:
I think it covers everything now.

@leo Could you have a final look?

dar5hak:
I'll push the docs tomorrow. Added that to the checklist.

lib/api.js
```diff
@@ -0,0 +1,22 @@
...
+const {spawn} = require('child_process')
+const path = require('path')
+
+// Packages
+const dargs = require('dargs')
+
+module.exports = (options = {}, directory = process.cwd()) => {
+  const scriptPath = path.join(__dirname, '..', 'bin', 'serve.js')
+  const aliases = {cors: 'o'}
+  options._ = [directory]  // Let dargs handle the directory argument
```
leo: Please add an empty line before this line.

lib/api.js
```diff
@@ -0,0 +1,22 @@
...
+
+  const cli = spawn('node',
```
leo: Please split it into multiple lines:

```js
const args = [
  scriptPath,
  ...dargs(options, {aliases})
]

const cli = spawn('node', args, {
  stdio: 'inherit'
})
```

package.json
```diff
@@ -53,6 +54,7 @@
     "boxen": "1.0.0",
     "chalk": "1.1.3",
     "copy-paste": "1.3.0",
+    "dargs": "^5.1.0",
```
leo: Please pin this dependency! => `5.1.0`

dar5hak:
Added docs. Please have a look at `readme.md`.

dar5hak:
Also, a small problem. I couldn't get the `ignore` argument to work. How do you invoke it on the CLI?

```bash
serve --ignore readme.md package.json ./
```

This fails with the statement: `Specified directory doesn't exist!`

leo:
@dar5hak `serve --ignore "readme.md,package.json"

tunnckoCore:
@leo @dar5hak why use comma-separated values? Why not just accept a glob patterns and pass them to some fast matcher like `nanomatch` or `micromatch`. Nanomatch has strict support for only globs, without exglobs and etc. And it is Bash spec compliant.

leo:
@tunnckoCore 

1. Not accepting comma-separated values and accepting glob values are two completely different things and aren't mutually exclusive.

2. If we're going to support glob values, you'll still need to support normal strings as well. Otherwise it won't be possible to ignore a single file.

3. If we're going to support glob values, we'll still use comma-separated values for that. Otherwise minimist will treat one glob value as the directory to serve. This means that we can just go on with this PR like I said [here](https://github.com/zeit/serve/pull/113#issuecomment-280906855).

😊

dar5hak:
@leo There. Everything works now. :tada:

leo:
@dar5hak Thanks a lot! 😊

PR #114, opened 2017-02-17
Promisify copy

disquisition:
`copy` needs to be promisified to expose errors from stderr within `copyToClipboard` (once xavi-/node-copy-paste#53 is merged this will be necessary to expose the error raised when the command required to copy to clipboard is missing).


leo:
Thanks! 😊

PR #116, opened 2017-02-18
Add opn module to open localURL in default browser

Fenwil:


leo:
I'd like to keep it like it is 😊

sylvainpolletvillard:
Would you consider adding it as an option ? I'm also interested by this

leo:
I would accept a PR for piping the URL to the command: `serve | open` or something similar.

PR #129, opened 2017-03-09
Alias boolean cli flags in minimist config

### LINKED ISSUE #128: arguments: -s does not do the same thing as --single
to reproduce: `serve -s dir` where dir contains an index.html 
Try visiting `localhost:5000/fakeurl`
You will receive a `not found` 404 rather than index.html

Repeat this with `serve --single dir` and the expected result (contents of index.html) are served.

The `-s` argument is ignored while `--single` works.

disquisition:
Looks like `minimist` defaults boolean flags to `false` which in turn messes with alias reconciliation in `args`. Simplest way to deal with this right now is to add `minimist` configuration for boolean flag aliases.

Closes #128

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -42,6 +42,13 @@ args
 
 const flags = args.parse(process.argv, {
   minimist: {
+    alias: {
+      a: 'auth',
+      C: 'cors',
+      S: 'silent',
+      s: 'single',
+      u: 'unzipped'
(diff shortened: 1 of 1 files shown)


leo:
Thanks!

leo:
[Released!](https://github.com/zeit/serve/releases/tag/5.0.2)

PR #130, opened 2017-03-11
Use clipboardy

floatdrop:
It uses `xsel` which is more reliable and [bundled](https://github.com/sindresorhus/clipboardy/pull/6)  with package to get things working on linux out of the box.

### DIFF
--- a/lib/listening.js
+++ b/lib/listening.js
@@ -1,12 +1,10 @@
 // Packages
-const {copy} = require('copy-paste')
+const {write: copy} = require('clipboardy')
 const ip = require('ip')
 const pathType = require('path-type')
 const chalk = require('chalk')
 const boxen = require('boxen')
-const {coroutine, promisify} = require('bluebird')
-
-const copyAsync = promisify(copy)
+const {coroutine} = require('bluebird')
 
 module.exports = coroutine(function * (server, current, inUse) {
   const details = server.address()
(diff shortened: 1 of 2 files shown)


lib/listening.js
```diff
@@ -1,5 +1,5 @@
 // Packages
-const {copy} = require('copy-paste')
+const {write: copy} = require('clipboardy')
```
disquisition: `clipboardy.write` returns a promise already so the `promisify` call below is unnecessary.

AndyBitz:
This will fail on bash on windows if `xsel ` is installed.

image

floatdrop:
@AndyBitz can you share the setup? Is it cygwin? What does `xsel` returns, if you launch it manually?

AndyBitz:
I'm using `bash on ubuntu on windows ` which is a linux subsystem on windows.

xsel returns
`xsel: Can't open display: (null)
: Success `

floatdrop:
@AndyBitz is calling `DISPLAY=:0 xsel` works for you?

AndyBitz:
@floatdrop returns
`xsel: Can't open display: (null) `
`: Connection refused`

floatdrop:
@AndyBitz patch to `execa` is ready, but can you check the exit code of `xsel` – is it non-zero?

AndyBitz:
@floatdrop yes it is 1

floatdrop:
@AndyBitz patch released as `0.6.1` – can you confirm, that issue gone?

AndyBitz:
@floatdrop Perfect! Everything works fine!

leo:
Sweet! 😏

PR #132, opened 2017-03-11
Don't freeze if path doesn't exist

### LINKED ISSUE #131: serve -s [bad dir] freezes
```
npm i -g serve@5.0.2
serve -s badpath
```

Expected: it tells me `badpath` ain’t there.
Actual: nothing happens.

screen shot 2017-03-11 at 9 41 13 pm

Node 7.7.2.

AndyBitz:
This closes #131.

`yield pathType.dir(current) ` and `pathType.dirSync(current) ` just return on `SIGINT `.
`fs.existsSync ` should be more reliable.


floatdrop:
@AndyBitz `path-type` also used in [lib/render.js](https://github.com/zeit/serve/blob/master/lib/render.js#L5) and [lib/server.js](https://github.com/zeit/serve/blob/master/lib/server.js#L10). Should it be replaced there too (and removed from `package.json`)?

AndyBitz:
@floatdrop I tried to change it there too. It didn't work as expected.
Also I figured that `fs.existsSync ` doesn't check if it is a directory or not, so I changed it to `lstatSync `.

leo:
Thanks a lot! 😏

PR #141, opened 2017-03-23
add --noclipboard flag

williamboman:


bin/serve.js
```diff
@@ -39,6 +39,7 @@ args
   .option('auth', 'Serve behind basic auth')
   .option('cors', 'Setup * CORS headers to allow requests from any origin', false)
   .option('silent', `Don't log anything to the console`)
+  .option('noclipboard', `Don't copy address to clipboard`, false)
```
williamboman: `args` doesn't seem to support "multi word" options, such as `--no-clipboard`.

leo: Can you make `args` support these?

williamboman: Sorry, I don't think I have time for that 😔

PR #152, opened 2017-03-29
solve icons with svgs in css. Abandon font-awesome

vedam:
hey all,
first of all thx for all of your great zeit-awesomeness.
I've wondered why to load the whole font-awesome-lib for just two icons (file and folder) and tried to solve that with svgs in css.

Would save about 75KB.

before

after

Hopefully you'll find it useful.
Just an idea.

### DIFF
--- a/assets/styles.css
+++ b/assets/styles.css
@@ -72,12 +72,12 @@ ul a {
   text-overflow: ellipsis;
 }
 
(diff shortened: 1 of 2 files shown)


leo:
WOW!

I just tested it with a directory and the reponse size shrinked from 100KB to 4.4KB.

Wonderful. Thanks! 😲

PR #166, opened 2017-04-19
default mime type text/html for SPA

JamesKyburz:
When refreshing a page in a SPA the page source is returned as text/plain.

This PR changes the content type to text/html if the single flag is set.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -141,9 +141,9 @@ module.exports = coroutine(function*(req, res, flags, current, ignoredFiles) {
     return stream(req, indexPath, streamOptions).pipe(res)
   }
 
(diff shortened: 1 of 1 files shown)


leo:
Why? What about the JS and CSS assets and vectors, etc.? Aren't they better suited with `text/plain` than with `text/html`?

JamesKyburz:
The assets are returned with the correct mime types. This is because the mime module uses their extensions.

The problem is when the url has no extension, so for example a url in a SPA with /awesome/cats would return text/plain and the html source. In a SPA the default page is index.html.

leo:
:ok_hand:

JamesKyburz:
Thanks!

PR #167, opened 2017-04-19
replace fs sync with async yield

JamesKyburz:
I noticed some sync method calls to fs/fs-promise which unless I am missing something block the event loop for each http request.

This PR replaces them with async yield.


leo:
Sweet, thanks!

PR #190, opened 2017-05-12
Add support for long-term caching of files with content-hashed filenames

jjavery:
Would this be useful to anyone else?

ETags are great but still require an If-None-Match/304 Not Modified request/response cycle. Tools like webpack are often configured to produce static assets with content-hashed filenames. The filename contains a hash of the file content, so when the content changes, the filename changes. For example, Create React App produces (via webpack) filenames like this:  `/static/js/main.54d5d5d6.js`. 

This pull request adds the boolean `--hash` option which enables detection of content-hashed filenames and serves the files with a max-age of 365 days. It also adds the `--regexp` option which can be used to change the default RegExp. The default RegExp matches the hash format used by Create React App.

https://webpack.js.org/guides/caching/#components/sidebar/sidebar.jsx

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -39,6 +39,8 @@ args
     'Time in milliseconds for caching files in the browser',
     3600
   )
+  .option(['H', 'hash'], 'Long-cache URLs with embedded hashes', false)
+  .option('regexp', 'Long-cache hash RegExp', '\\.[0-9a-f]{8}\\.')
   .option('single', 'Serve single page apps with only one index.html')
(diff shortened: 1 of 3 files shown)


leo:
Thank you for the effort you've put into his! But please consider creating [an issue](https://github.com/zeit/serve/issues) the next time before you start tackling the topic, so that we can discuss it and agree on it. Because in this case, I don't think it's a good idea to merge this into the project, since we're trying to keep the core as small as possible.

PR #192, opened 2017-05-13
Closes zeit/serve#159

jjavery:
Closes zeit/serve#159

Shut down server on both SIGTERM and SIGINT to support Docker as `docker stop` sends SIGTERM to PID 1.

Be sure to use the exec form of the CMD instruction to get serve on PID 1:

`CMD [ "serve", "-s", "build", "-p", "80" ]`

The shell form of the CMD instruction puts serve on PID 2, which won't work:

`CMD serve -s build -p 80`

See this for more info: https://www.ctl.io/developers/blog/post/gracefully-stopping-docker-containers/


leo:
The tests are failing. Can you please fix them?

PR #198, opened 2017-05-18
Open local address in default browser

webtaculars:
Open local address in default browser by using option --local.
`serve --local` 

**EDIT**
Command is now `serve --open`


leo:
👍 I'd rather call it `--open`

webtaculars:
Yeah 😀. It should be `--open'. I will update it.

webtaculars:
@leo I have changed `--local` to `--open`. Need to update anything else?

lib/listening.js
```diff
@@ -67,6 +68,13 @@ module.exports = coroutine(function*(server, current, inUse, clipboard) {
       } catch (err) {}
     }
 
+    if(open) {
```
leo: Please add a space before the `(`

lib/listening.js
```diff
@@ -67,6 +68,13 @@ module.exports = coroutine(function*(server, current, inUse, clipboard) {
...
+      try {
+        opn(localURL)
+        message += `\n\n${chalk.grey('Local address has opened in browser!')}`
```
leo: Let's not show any additional messages in the terminal!

lib/listening.js
leo: Also please test for `isTTY`

bin/serve.js
```diff
@@ -108,7 +109,7 @@ detect(port).then(open => {
   server.listen(
     port,
     coroutine(function*() {
-      yield listening(server, current, inUse, flags.noClipboard !== true)
+      yield listening(server, current, inUse, flags.noClipboard !== true, flags.open === true)
```
leo: Why the comparison? `flags.open` should be enough

webtaculars:
@leo Updated all the changes you asked.

bin/serve.js
```diff
@@ -50,7 +50,7 @@ args
   )
   .option('silent', `Don't log anything to the console`)
   .option('no-clipboard', `Don't copy address to clipboard`, false)
-
+  .option('open', `Open local address in browser`, false)
```
leo: Use ' instead of ` here, please

webtaculars: should I change it for silent and no-clipboard also?

leo: @webtaculars No, because they have a ' in the value.

webtaculars: we can escape it using backslash

leo: Yes, but prettier might revert that. Try it!

leo:
Thanks! :v:

PR #209, opened 2017-05-31
Add default caching options when serving with single flag

### LINKED ISSUE #208: Add flag for serving SPAs
Would be great to a flag for SPAs that adds two features

- [x] Changes caching options so that index.html has a cache of 0, and all other files have the amount passed in the --cache flag (and maybe a sensible default if not cache flag was passed)
- [x] Changes the folder redirect rule so that all files that aren't found get redirected back to /index.html

Happy to make PR for this.

ro-savage:
Closes #208

For deploys with the `single` flag, it changes the `cache-control max-age` to `0` for `index.html`, and 1 day for everything else.  

You can still overwrite the max-age with the `cache` flag, even when deploying with `single` flag, however it will keep index.html at `max-age 0`

I also removed the defaulting of caching being 3600ms (approx 3seconds) as that seemed pointless. Instead just lets micro decide on the default?

This is an alternative to #190. Either works, but would like to see caching done better for SPAs :)

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -34,12 +34,11 @@ if (process.env.NODE_ENV !== 'production' && pkg.dist) {
 
 args
   .option('port', 'Port to listen on', process.env.PORT || 5000)
+  .option('cache', 'Time in milliseconds for caching files in the browser')
   .option(
-    'cache',
-    'Time in milliseconds for caching files in the browser',
-    3600
+    'single',
+    'Serve single page apps with only one index.html. Sets cache for 1 day.'
   )
-  .option('single', 'Serve single page apps with only one index.html')
   .option('unzipped', 'Disable GZIP compression')
   .option('ignore', 'Files and directories to ignore')
   .option('auth', 'Serve behind basic auth')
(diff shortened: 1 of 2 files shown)


leo:
Why should we set a different max-age for index.html? 😊

ro-savage:
Because the vast majority of SPAs use a cache busting technique where each bundled JS/CSS has a unique hash (or hash based on content). However cache busting won't work if `index.html` is cached because the user will see the old index.html that points to the old bundles.

This allows you, for example, to cache your hashed JS/CSS files for a year but still have the site instantly update when a change is made.

leo:
@ro-savage I thought browsers don't cache HTML files?

ro-savage:
@leo - Nope, browsers will respect cache unless you do a refresh. e.g. press the refresh button, press enter a 2nd time in the url bar, cmd+R)

It can be confusing when testing the behaviour, it just confused me when I was double check this!

When you press enter on the URL bar that forces the browser to refresh (and make a request with max-age: 0). 

So if you make a change, then press 'enter' it will refresh and you will get an uncached copy. But if you just type the URL and go to the site normally, you'll get a cached copy.

Heres a quick video example, using now and serve.
https://www.youtube.com/watch?v=bqwDREbSbdw

leo:
Looking at this PR, I think we don't have a default caching value anymore? We need that.

ro-savage:
As mentioned in the original comment:

> I also removed the defaulting of caching being 3600ms (approx 3seconds) as that seemed pointless. Instead just lets micro decide on the default?

If we need a default, maybe either 0 seconds, aimed as people using serve in development. Or 1 day (or more?) aimed at people using serve in production.

3600ms seems like a strange default to me. But happy to add in 3600ms default if that's what you'd like?

leo:
Thanks for the effort you've put into this PR, but I had to implement a custom solution (the branch had merge conflicts so I thought it would be faster). 🙏

PR #218, opened 2017-06-08
Fixed custom 404 pages

rosszurowski:
This PR fixes an issue with `serve v5.2.2` where a custom 404 page (ie. `404.html`) wasn't being sent.

From my triaging, it seems like the issue is with [this line](https://github.com/zeit/serve/blob/8d556096d8ccca2409ee3b0d30fc3ac862a1c08b/lib/server.js#L86), which checks if `flags.single === undefined`. Since the default value is actually `false` this block never gets triggered and we never end up sending the custom 404 page.

This change doesn't _seem_ to introduce regressions even with different options, but it's hard to actually tell without automated tests 😢

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -83,7 +83,7 @@ module.exports = coroutine(function*(req, res, flags, current, ignoredFiles) {
   }
 
   const relatedExists = yield fs.exists(related)
-  if (!relatedExists && flags.single === undefined) {
+  if (!relatedExists && !flags.single) {
     return micro.send(res, 404, notFoundResponse)
   }
 


DonnieWest:
@leo Is there anything I can do to help get this merged in? This seems like it'll fix my custom 404 page issues and also my service worker on my website

leo:
The problem is that we need to ensure that [this](https://send-patch.now.sh) doesn't happen again. I've recently merged [this](https://github.com/zeit/serve/commit/ade5a147901bac4ffb7d59835e4c89955ee2044a) and it brought the problem back after it was fixed. Since your PR looked similar, I thought it might create the same problem again.

rosszurowski:
@leo What exactly is that issue? I can't find any details about it other than that:

> The issue was associated with the usage of the send middleware without supplying a root option.

If you let me know more (eg. what kind of request can cause the issue, how it's related to the 404 page), I'm happy to manually test this PR to ensure that there's no regression.

Also, would you be open to a PR introducing automated testing? Seems like a good way to ensure that kind of regression never happens again.

d4rky-pl:
From what I see there's exactly the same code few lines below - https://github.com/zeit/serve/blob/master/lib/server.js#L138 - so one way or another there's a bug!

I've been trying to figure out what the security issue is and how that change could reintroduce it. [From what I see in next.js changes in 2.4.1 version](https://github.com/zeit/next.js/commit/43c447edd0c23db5a9acc477befa77a42826d1a3) there is [only one change related to security](https://github.com/zeit/next.js/commit/43c447edd0c23db5a9acc477befa77a42826d1a3) and I'm afraid I'm not that knowledgeable in internals of both next.js and this library to guess how those two are connected :( 

@leo it would be massively helpful if you provided more info on what the security issue in this case was and if there's some way of mitigating it. I'm happy to dive into it and provide necessary patch but I need more info, POC, description, anything :)

leo:
Sweet! I checked it manually and everything seems to work correctly. Thanks! 😊

PR #223, opened 2017-06-14
add a new json flag to list directory as json content

maxired:
Hi,

this is an attempt at adding a new flag, `json`, which would allow serving a directory listing, not with the current hbs template, but using a JSON format.

This is pretty useful at a tool for development purpose.

Let me know what you think about it.

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -40,6 +40,7 @@ args
     3600
   )
   .option('single', 'Serve single page apps with only one index.html')
+  .option('json', 'Use JSON to list directories')
   .option('unzipped', 'Disable GZIP compression')
   .option('ignore', 'Files and directories to ignore')
   .option('auth', 'Serve behind basic auth')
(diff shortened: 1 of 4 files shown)


leo:
How is this useful? Thanks!

maxired:
While the current directory listing is done for Humans, this implements a directory listing for Machines.

Let's say I am building a UI listing json files in a folder and displaying them. Such an options will prove to be really usefull.

albinekb:
Better to follow HTTP standards and look for the `Accept` header, and if it's `appliction/json` send it in json.

maxired:
@albinekb I was originally thinking the same, but then I thought that the impact on the code base (that I don't really know), would be bigger.

leo:
Thank you for the effort you've put into his! But please consider creating [an issue](https://github.com/zeit/serve/issues) the next time before you start tackling the topic, so that we can discuss it and agree on it. Because in this case, I don't think it's a good idea to merge this into the project, since we're trying to keep the core as small as possible.

I'm sure there are other packages on npm that do this. And even if not, it should be fairly easy to write.

maxired:
Hey @leo, thanks for the detail anwser.

I totally understand your point, but I am happy I handle it this way  so it suits the use case I faced, and hopefully somebody else will find this pr/fork when looking for something similar. 

When you say that you are  trying to keep the `core` pretty small, do you mean there is an easy  way to add some custom behavior to a serve powered server?

leo:
@maxired Well, I honestly don't think `serve` is suitable for such situations. Especially as it renders files when you're trying to access them - instead of the directory json listing.

Here's a basic example with node's native API:

```js
const {readdirSync} = require('fs')
const files = readdirSync('<your-directory>')

JSON.stringify(files)
```

It's as easy as that. Now you only need to wrap this into `http` and modify the array a little.

Issue #239, opened 2017-06-30
Serving! (on port 13302, because 13302 is already in use)


danielking:
serve always show message like this because of unnecessary strict equality comparison.

danielking:
i created a pr, #240

leo:
https://github.com/zeit/serve/pull/240

PR #240, opened 2017-06-30
Update serve.js

danielking:
change unnecessary strict equality comparison to abstract equality comparison


danielking:
#239

leo:
Is one of those of a different type? If so, we need to convert it to an integer, because the linting doesn't allow non-strict comparisions.

leo:
https://github.com/zeit/serve/pull/242

Issue #224, opened 2017-06-15
Explicit cache bust


KrekkieD:
Would it be possible to add a param that explicitly sets no-cache headers? 

Reason being is I use `serve` in a development deployment scenario for reviewing purposes. The people that review the app are not tech savvy so I have to remind them every time to refresh the page properly. Next to that, some browsers have very sticky default caching behaviour, which makes it even worse.

I'm currently serving with `serve -s -c 0` which does set the max-age to zero, but that doesn't seem to be behaving the same as no-cache headers.

leo:
We should set the "no-cache" one if `-c` is `0` maybe?

leo:
Fixed in https://github.com/zeit/serve/pull/245

jespertheend:
is no longer working?

PR #245, opened 2017-07-07
Disable cache when `--cache` is set to `0`.

fabe:
This implements #224 and sets a `Cache-Control: no-cache` header when serving with `serve -c 0` or `serve -s -c 0`. Had to manually set it because `send` doesn't allow string values for `cacheControl`.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -95,6 +95,11 @@ module.exports = coroutine(function*(req, res, flags, current, ignoredFiles) {
 
   if (flags.cache) {
     streamOptions.maxAge = flags.cache
+  } else if (flags.cache === 0) {
+    // Disable the cache control by `send`, as there's no support for `no-cache`.
+    // Set header manually.
+    streamOptions.cacheControl = false
+    res.setHeader('Cache-Control', 'no-cache')
   } else if (flags.single) {
     // Cache assets of single page applications for a day.
     // Later in the code, we'll define that `index.html` never


leo:
Wonderful, thanks!

monolithed:
How to disable cache now? It seems these changes were removed? `--debug?`

jespertheend:
adding a `serve.json` file to my server root with this json worked for me
```json
{
  "headers": [
    {
      "source" : "**",
      "headers" : [{
        "key" : "Cache-Control",
        "value" : "no-cache"
      }]
    }
  ]
}
```

specious:
I'm using serve 12.0.1 and this configuration doesn't seem to be working.

specious:
I'm using serve 12.0.1 and this configuration doesn't seem to be working.

The server responds with header:
```
Cache-Control: max-age=0
```

However, it also responds with HTTP status `304 Not Modified`.

damianobarbati:
@jespertheend is that still working for you? I see no cache headers being set at all with:
```json
{
  "headers": [
    {
      "source" : "*.html",
      "headers" : [{
        "key" : "Cache-Control",
        "value" : "no-store"
      }]
    },
    {
      "source" : "*.(js|css|map|jpg|jpeg|gif|png|ico)",
      "headers" : [{
        "key" : "Cache-Control",
        "value" : "public; max-age=86400"
      }]
    }
  ]
}
```

jespertheend:
@damianobarbati I haven't used serve in a while to be honest, but on one of my old projects it seems to still be working. Though it's using serve 10.0.1

PR #255, opened 2017-07-23
fix missing zero in one day cache number

albertstill:
86400000 milliseconds is 1 day

https://www.google.com.au/search?q=86400000+milliseconds+to+days&spell=1&sa=X&ved=0ahUKEwiTmITf1Z7VAhXMpJQKHUojDWwQvwUIIygA&biw=1440&bih=776

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -104,7 +104,7 @@ module.exports = coroutine(function*(req, res, flags, current, ignoredFiles) {
     // Cache assets of single page applications for a day.
     // Later in the code, we'll define that `index.html` never
     // gets cached!
-    streamOptions.maxAge = 8640000
+    streamOptions.maxAge = 86400000
   }
 
   // Check if directory


leo:
Thank you! But the MDN documentation states that it should be [seconds](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control).

albertstill:
I know the Cache-Control HTTP header is in seconds, but this is a `send` options object, which is in [milliseconds](https://github.com/pillarjs/send#maxage). Right now it's missing a single zero, so it equates to 2.4 hours, and your API is saying it should be [24 hours](https://github.com/zeit/serve/blob/dae446eb35480cb1db3fedd7e37983378093945c/lib/options.js#L16).

leo:
Oh,

leo:
[Released!](https://github.com/zeit/serve/releases/tag/6.0.3)

albertstill:
cool thanks

PR #269, opened 2017-09-29
Fixed caching of SPAs and `/index.html`

MarcoThePoro:
See https://github.com/zeit/serve/issues/258

Previously, `serve` with `--single` would only block caching for paths that don't correspond to any existing files or directories. 

It would not block caching for paths such as the following, even though the SPA index.html is served:
- `/`
- `/index.html`
- `/some-directory/` (where `/some-directory` exists but `/some-directory/index.html` does not)

These changes should ensure that caching always gets blocked when serving `/index.html` in SPA mode.


leo:
Thank you! 😊

PR #270, opened 2017-10-01
bump dependencies, primarily `send` for mime time upgrades

rauchg:


rauchg:
I verified the changelogs of `send`, should be a pretty smooth upgrade.

PR #274, opened 2017-10-11
Support HTTPS

### LINKED ISSUE #178: Option for enabling HTTPS
Any way to do it or is it planned?

justanotherdot:
This should hopefully resolve #178. It uses the example given in the
`micro` examples of `with-https`. It has some rough edges still that
needed to be sanded down:

* Choosing a specific short flag name for ssl (and if the initial flag
should still be called ssl; I'm tempted to rename it to `https` or what
have you) - would like if someone chimed in about their design thoughts for this.

* The example always acks with 200 and the example object - should just be a matter of passing off the handler as `fn` to `microHttps`.

* It does not support compression and merely overwrites the previous
server var - should also be trivial, just compress the handler before handing it off to whichever server will take it.


justanotherdot:
I should be able to get to  last two bullet points without much delay; the first needs someone to chime in on design wise. :slightly_smiling_face:

justanotherdot:
Last commit should resolve the latter two bullet points; just need to know if we should change the flag name? (I'm going to vote 'yes'! Maybe `https`?)

justanotherdot:
Also just realized there should probably be:

* a check for at least min of node v7.6 to support the `async` keyword unless it's unnecessary.
* a way to pass the cert + key manually if one is inclined to do so.

Looks like [`http-serve`](https://github.com/indexzero/http-server) has `--ssl` and `-S` for the flags they use, so it might be ok to keep it and introduce `-S` here for consistency.

leo:
wow, this looks awesome! I added a few tiny comments.

lib/options.js
```diff
@@ -50,6 +50,11 @@ exports.options = [
     name: 'treeless',
     description: `Don't display statics tree`,
     defaultValue: false
+  },
+  {
+    name: 'ssl',
+    description: 'Serve with ssl',
```
leo: > Serve content using SSL

bin/serve.js
```diff
@@ -67,7 +73,17 @@ const handler = coroutine(function*(req, res) {
   yield serverHandler(req, res, flags, current, ignoredFiles)
 })
 
-const server = flags.unzipped ? micro(handler) : micro(compress(handler))
+const httpsOpts = {
+  key: cert.key,
+  cert: cert.cert,
+  passphrase: cert.passphrase
+}
```
leo: Please add a newline after this line 😊

bin/serve.js
```diff
@@ -4,6 +4,7 @@
 const path = require('path')
 
 // Packages
+const https = require('https')
```
leo: This is not a package - it needs to be part of the "Native" import group.

justanotherdot: Sure thing; force of habit to add to the end and glossed over the comments :slightly_smiling_face:

lib/options.js
justanotherdot: Sounds good; wasn't sure what the ideal copy was so I figured someone would eventually chime in :wink:

justanotherdot:
Thanks! I'll be sure to run `npm` to update the`package-lock.json` :smile:

justanotherdot:
Per `lib/server.js` conflicts, I think the global version of `prettier` I have installed may reintroducing that. I'll resolve here to circumvent it running.

Per `package-lock.json` I'll just resolve to what master has for now.

justanotherdot:
Should be good, and should also be using the appropriate non-master (`npm install`) `package-lock.json`. :+1:

bin/serve.js
justanotherdot: Done :smile:

leo:
Thank you so much! I'll make sure to update the snapshots and lockfiles to be really sure.

albinekb:
:nice: use of the micro example 👍

justanotherdot:
No worries, and thanks for writing that example, @albinekb! Make hacking up the diff a snap :ok_hand:

albinekb:
OSS FTW! 💯

jadbox:
This needs documentation in the README.

paullaffitte:
Does this feature has been disabled? The `--ssl` flag doesn't exists anymore, and neither nothing related to it in the `--help` flag.

devinrhode2:
Indeed need documentation! I don't know what to do

nermin99:
@justanotherdot What happened to your flag? The `--ssl` flag doesn't work anymore and there's only `--ssl-cert` and `--ssl-key`.

justanotherdot:
I am unsure I’m afraid. I was not responsible for the change removed the
flag. Looks like those new flags made their way in over at
https://github.com/vercel/serve/pull/520

nermin99:
@leo why was support for automatic HTTPS removed? As per #520 and #511 it is understandable that users might want to use manual certs and keys, **but** there is still, no even more, use cases for automatic SSL/HTTPS. One of the biggest reasons why `serve` is so popular is because it's a one-line command to literally "serve" your application. If a user wants or for some reason needs HTTPS they don't wanna have to go through a whole process of installing another tool just to get it.

Personally I stumbled upon serve when building my create-react-app and wanted to "serve" it. But because I use service workers and push notifications I need to serve it over HTTPS so I can test on other devices on the network. Sure I can get HTTPS for my self as per #520, but my co-workers can't, since they might not have their own manual certs and keys... you see the problem?

There is no reason why both manual and automatic HTTPS can't be supported.

Issue #233, opened 2017-06-24
Introduce tests


leo:
We need automated tests to ensure that the code always works correctly!

MarkRabey:
I'd love to help out with this!

leo:
@MarkRabey Wonderful! I'd love to accept a pull request.

MarkRabey:
@leo great, I'll start with just adding a testing framework and setting it up. Any preferences? I was thinking [Mocha](https://mochajs.org/).

leo:
@MarkRabey AVA, please 😊

leo:
✅ thanks @maxbeatty

PR #275, opened 2017-10-12
Introduce tests

maxbeatty:
ref #233 

adds a few tests using Ava and NYC like other ZEIT projects


leo:
Awesome, thank you! 😊

Issue #237, opened 2017-06-29
Improve listing files with long name


iwazaru:
I'm using **serve** to host and share files that may have long names. Current presentation only displays the first characters of the file's name in most cases, which can lead to confusion.

![Listing files](https://user-images.githubusercontent.com/5627688/27691117-63fe3210-5ce3-11e7-94f1-610e4d7a9bb6.png)

I would love to have an option that would allow displaying one file per line, with the full name of the file shown.

tohjustin:
@leo I think this should be a relatively straightforward change, should I just add a button to toggle the layout? Also, do you have any mock ups that I could work with? =0

leo:
Fixed in https://github.com/zeit/serve/pull/276

sunlei4076:
not work in ios device

PR #276, opened 2017-10-15
Added ability to toggle the list view

barinali:
Hello,

First of all, I'd like to thank you for `serve`. So thanks.

I have built a feature that provides toggling the view between single column view and multiple columns view, regarding issue #237.

Here is a quick preview which is unfortunately poor quality;
gif

I'm open to apply any suggestion here if any.

### DIFF
--- a/assets/styles.css
+++ b/assets/styles.css
@@ -10,6 +10,25 @@ main {
   max-width: 920px;
 }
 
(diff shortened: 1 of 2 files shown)


leo:
Wonderful contribution - thank you!

Issue #257, opened 2017-08-16
Suggestion: SPA should 404 on fetching obvious non-html assets


transitive-bullshit:
For SPAs, a common issue I've run into is that as hashed static resources change between deployments, a stale client requests an old asset (oftentimes due to a client-side service worker), and failing to find the static resource, `serve` naively returns `index.html` which is almost never what you want and leads to weird client-side errors.

See [here](https://github.com/PolymerElements/app-storage/issues/81) and [here](https://github.com/react-boilerplate/react-boilerplate/issues/1676) for solid descriptions of the general problem aside from `serve`.

I realize this is how most static file serving solutions work, but I know we can do better for the majority use case, where the requested resource is in a blacklist of common content-types (e.g., `js`, `css`, `png`, `jpg`, `svg`, etc).

I'd like to propose adding an optional flag to augment SPA mode whereby these types of resource requests return 404 if not found, instead of naively returning `index.html`.

Thoughts?

sergiodxa:
In a SPA your 404 is usually being rendered client side, that means you need to first get your `index.html`, start your app and then check the URL to render your app or render a 404.

fisch0920:
@sergiodxa the whole point here is that the server isn't returning a 404 -- it's returning `index.html` for a missing static asset as opposed to returning a 404 which could be gracefully handled client-side.

leo:
Thank you for the effort you've put into this!

This is now possible by supplying a `serve.json` file. All thanks to `serve-handler`, as described in the notes for the [new release](https://github.com/zeit/serve/releases/tag/7.0.0).

Simply add `headers` to your config file, as described in the `serve-handler` [documentation](https://github.com/zeit/serve-handler#headers-array).

minhdao88:
Sorry to bother but I just stumbled upon this issue recently, and after checked serve-handler documentation I still don't understand how to config it to return 404 status instead of the HTML file. 

Any help would be greatly appreciated.

PR #279, opened 2017-10-23
Added HTTP 404 to non-existent files in SPA mode

barinali:
Hello,

I have seen issue #257 and wanted to contribute again. I think it would be better if we return HTTP 404 for non-existent files in SPA mode. Also, I believe that it shouldn't be a problem for SPAs since there is nothing to be crawled regarding to non-existent files by bots.

I am happy to apply any other approach if there is.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -168,11 +168,16 @@ module.exports = coroutine(function*(req, res, flags, current, ignoredFiles) {
   // - The path matches exactly `/index.html`
   if (
     flags.single &&
-      (!(yield fs.exists(related)) ||
+    (!(yield fs.exists(related)) ||
       related === path.join(current, '/index.html'))
   ) {
     // Don't cache the `index.html` file for single page applications
     streamOptions.maxAge = 0
+
+    // Set HTTP 404 since there is not an actual file.
+    if (pathname !== '/index.html') {
+      res.statusCode = 404
+    }
 
     // Load `index.html` and send it to the client
     const indexPath = path.join(current, '/index.html')


leo:
Thank you the effort you've put into this, but this doesn't make sense. 😅 All request always need to be routed to the index file, because SPAs will handle them client side.

barinali:
@leo oh, it's routed to `index.html` anyway. There is no change about that. The thing is that; we're serving index.html with HTTP 404 if there is no such an actual file. In this case, if served SPA is using a service worker, the service worker can handle replaced assets well. Otherwise, when service worker tries to fetch them, it fetches them as HTTP 200 which seems fine to SW.

leo:
But shouldn't it be `pathname === '/index.html'` instead then? Because right now, you're applying that 404 error if the file is not `index.html` AND it doesn't exist.

barinali:
Let me explain it a little bit in detail;

if the end-user;
- tries to reach`/` or `/index.html`, `index.html` is being served with HTTP 200.
- tries to reach a file that exists on the folder, the requested file is being served with HTTP 200.
- tried to reach a URL that doesn't exist on the folder, `index.html` is being served with HTTP 404 to notice the browser (or developer) that we're just serving index.html for SPA purpose. Then that non-existent file can be caught by service worker or whatever else.

leo:
Not sure if we should really respond with 404. It makes the browser think the resource was not found, although it was actually found (but on the client)...

barinali:
Service workers are commonly used in SPAs. So I believe that we need to support this 404 approach somehow as it's requested in the issue. But it's up to you. As far as I know, there is nothing wrong in this approach regarding to search engines (since the developers serve their SPA via something like Preprender, and [Prerender](https://prerender.io/documentation/best-practices) can decide responding it with HTTP 200 or 404 by a meta tag)

rauchg:
I think 404 associated with an actual response is fine. It'd be up to that response to _handle_ the 404 and render something. What would be weird is sending 404 and actually rendering content.

barinali:
But we're talking about single page applications here. I couldn't get what is wrong with this approach though. Because on this branch, if you have an SPA, and a route like `/profile`, we'll serve `/profile` page with index.html, right? But it would be being served with HTTP 404 since there is no such a `profile` file on the directory.

Besides that, if anyone wants to render a real 404 view, they should do it on their SPA level.

In my opinion, it shouldn't matter if the browser sees `/profile` page with HTTP 404, or HTTP 200 actually. Because it's just a network activity that doesn't trigger or affect anything as far as I know for the browser. HTTP Status of the page becomes meaningful when we write a code block regarding that, like service worker case.

I think we should pay attention to service workers as much as we care about SPAs. So I'm still open to any suggestions. I may refactor the code in such a way that responds with index.html and HTTP 404 for non-existent files (with extensions like main.blabla.js, style.lol.css) instead of responding index.html and HTTP 404 for all non-existent URLs if this makes sense to you all.

leo:
We're working on deprecating the `--single` flag because it's basically just a hack and will never cover all the possible configuration ways. In turn, we'll soon provide a config file for it with the [superstatic](https://github.com/firebase/superstatic) spec that lets you easily route requests around without typing much text.

PR #282, opened 2017-10-26
Added wildcard support to the `--ignore` option

### LINKED ISSUE #281: Ignore option can't match wildcards
I discovered using `serve` that `ignore` option can match only exact file names.
So I forked the repo and I played with this: I'd like to use wildcards, too.
I'd like to use the meta-character `*` as I would use it in bash, to ignore every .js file passing "\*.js" or to ignore any file whose name begins by 'A' passing "A*\".

I'm submitting a PR with my work: #307.

nzambello:
Fixes #281:
> I'd like to use the meta-character `*` as I would use it in bash, to ignore every .js file passing "*.js" or to ignore any file whose name begins by 'A' passing "A*".


nzambello:
What I submitted can obviously be reviewed and changed with reviews.

I'm not confident with "hidden" option's name.
I initially chose 'hidden', but I had conflicts with 'help', so I moved to 'noshow' but the same with '-n'.
I finally chose 'donotshow' because on '-d' there was not anything.

leo:
You can pass an array to args with the short and long name: `['s', 'single']`. Please use `--hide` with a different short flag that's available. Also `args` should do this by itself, so please try it without the array before you use it.

nzambello:
I tried with: `name: ['h', 'hide']`, but calling
```serve -h```
it shows the usage informations.
The same without using the array.
  
So I setted `hide` option to:
```    name: ['H', 'hide']```

It's ok @leo?

nzambello:
As reported in #283, I was making some mistakes. I apologize for this.
I fix the PR adding only the wildcard support.

nzambello:
@leo Any feedback? What about this PR?

Issue #285, opened 2017-10-28
Use image SVG icon for files that are being served as images (.gif, .jpg, .png, .svg)


tohjustin:
From the designs found in https://github.com/zeit/art/blob/master/serve/look.sketch
<img width="1392" alt="sketch-preview" src="https://user-images.githubusercontent.com/8844886/32138910-a349c9cc-bbf0-11e7-9075-569e35445307.png">

Currently `serve` uses the same icons for files & images, would be great if we have a different icon for images (similar to the one shown in the screenshot above).

leo:
https://github.com/zeit/serve/pull/286

PR #286, opened 2017-10-28
Added icons for image files (#285)

tohjustin:
Currently images & files have the same icon:
before-change

With this change, we will have a fancy icon to differentiate images (.gif, .jpg, .png, .svg) from other types of files:
after-change


leo:
Thank you!

Issue #267, opened 2017-09-19
middleware support


Kikobeats:
This issue continues https://github.com/zeit/micro/issues/225 conversation about how to use `micro` + `serve`.

In my case, I want to server an static file (`index.html` with `js` and `css`) under  no query params, like a main entry page to the API.

Expected API inspired in [serve-static](https://www.npmjs.com/package/serve-static):

```js
const {middleware} = require('server')
const servePublic = middleware(path.join(__dirname, 'public'))

module.exports = (req, res) => {
  res.send(servePublic)
}
```

kavikaran:
is there any option for swagger documentation of micro service ?

daliborgogic:
Here ya go, simple micro server:
```js
const url = require('url')
const fs = require('fs')
const path = require('path')
const { send } = require('micro')

const mimeTypes = {
  '.ico': 'image/x-icon',
  '.html': 'text/html',
  '.js': 'text/javascript',
  '.json': 'application/json',
  '.css': 'text/css'
}

module.exports = async (req, res) => {
  const parsedUrl = await url.parse(req.url)
  let pathname = `.${parsedUrl.pathname}`

  fs.exists(pathname, exist => {
    if (!exist) {
      send(res, 404, 'You fucked up')
      return
    }

    if (fs.statSync(pathname).isDirectory()) {
      pathname += '/index.html'
    }

    fs.readFile(pathname, (err, data) => {
      if(err){
        send(res, 500, 'I fucked up')
      } else {
        const ext = path.parse(pathname).ext
        res.setHeader('Content-type', mimeTypes[ext] || 'text/plain' )
        send(res, 200, data)
      }
    })
  })
}
```

kavikaran:
it serves only static html file.

i want to whole bundle of html,css,and js. for example we serve public directory in express nodejs framework.

```
const docs = cors(async(req, res) => {
   let file = path.join(__dirname, './public');

  fs.exists(file, exist => {
    if (!exist) {
      send(res, 404)
      return
    }

    if (fs.statSync(file).isDirectory()) {
      file += path.join(__dirname, './public');
    }

    fs.readFile(file, (err, data) => {
      if (err) {
        send(res, 500)
      } else {
        res.setHeader('Content-type', mime.getType(file))
        send(res, 200, data)
      }
    })
  })
})
```

leo:
Thank you for the effort you've put into this!

[...]

PR #288, opened 2017-10-29
Fixed parent directory not sorted to the front of the list (#267)

tohjustin:
**Proposed Fix**
1. Sort the array of files in the current directory.
2. Add the parent directory (if needed) to the head of the sorted array.

**Additional Note**
- Not sure it's a better idea to extract out the file sorting logic into a function so we can test it


leo:
Thank you! :blush: Also please make sure to not mention the related in the PR name, that's anti-pattern. Instead of that, use [keywords](https://help.github.com/articles/closing-issues-using-keywords/).

PR #298, opened 2017-11-28
Compare flags.cache to the string '0', not 0

jeffposnick:
The logic in https://github.com/zeit/serve/pull/245 is faulty, as `flags` is set via

```js
const flags = args.parse(process.argv, { minimist })
```

and that will assign a string representation of all of the flag values.

This PR updates the code to handle `-c 0` by doing the comparison against the string `'0'` instead of the number `0`, to account for the fact that `flags.cache` is a string.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -95,7 +95,7 @@ module.exports = coroutine(function*(req, res, flags, current, ignoredFiles) {
 
   if (flags.cache) {
     streamOptions.maxAge = flags.cache
-  } else if (flags.cache === 0) {
+  } else if (flags.cache === '0') {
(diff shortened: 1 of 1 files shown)


jeffposnick:
Actually, my mistake. `-c 0` should be working properly in `master` already.

I was thrown because of that `flags.cache === 0` check, which I don't think is every going to be `true`... but the initial check of `flags.cache` will be `true`, since `'0'` is a true value, so I don't think that other clause will ever be triggered.

So `maxAge=0` is being set, but that whole `Cache-Control: no-cache` introduced in https://github.com/zeit/serve/pull/245/files isn't being set. They have similar effects, though.

PR #304, opened 2017-12-17
Add `--markdown` option, for reading docs more friendly.

fimars:
The new feature, for reading docs friendly. Maybe not bad? 

image


leo:
Thanks for the effort you've put into this, I highly appreciate it!

It looks like it wasn't easy, so again: Thanks a lot! 🙏 However, I have to say that this kind of feature doesn't make sense to implement into `serve`, in my opinion. We can't have every package render its own documentation, that would be madness... 😅 

But again: Huge thanks! And please ensure to talk with us/community about something like this before putting the effort into it. Just create an issue first next time! 😊

fimars:
Thank you for your courtesy and advice! 🙏 Your point of view is very correct, I agree, apologize for their own recklessness. 😅

PR #305, opened 2017-12-19
Recycle output of call to fs

fritz-c:
There was already a variable set to the result of the call to `yield fs.exists(related)`. Rather than call the same file system operation twice, I recycled the result of the first call. It's a tiny optimization.


leo:
> It's a tiny optimization.

Saving us milliseconds. Thank you!

leo:
Just went out with the latest release!

fritz-c:
Thanks for the quick response!

PR #307, opened 2017-12-23
Added wildcard support to the `--ignore` option

### LINKED ISSUE #281: Ignore option can't match wildcards

nzambello:
Fixes #281:
> I'd like to use the meta-character `*` as I would use it in bash, to ignore every .js file passing "*.js" or to ignore any file whose name begins by 'A' passing "A*".


leo:
Thank you for the effort you've put into this!

This is now possible by supplying a `serve.json` file. All thanks to `serve-handler`, as described in the notes for the [new release](https://github.com/zeit/serve/releases/tag/7.0.0).

Simply use `rewrites` to forward the paths (supports globs and regular expresssions) of your choice to your 404 route.

PR #309, opened 2017-12-29
Fixed typo in option alias

just-boris:
Currently it interprets {cors:true} option as open:true, which is not expected by users.
Fixes https://github.com/zeit/serve/issues/296


leo:
Thank you!

just-boris:
@leo can you make a release? I want to stop depending from the forked version in my project.

leo:
[Done!](https://github.com/zeit/serve/releases/tag/6.4.4)

PR #313, opened 2018-01-15
Added "margin: auto" centering main content

### LINKED ISSUE #284: Max-width in files view
I have a question: why is there that `max-width: 920px` to `main`?
Isn't it better if it could take more space?

If it's better as it is, what about centering it with a `margin: auto`?

nzambello:
Fixes #284.

I prefer main content to be at the center of the page. If you prefer it also, add this line of CSS.

### DIFF
--- a/assets/styles.css
+++ b/assets/styles.css
@@ -8,6 +8,7 @@ body {
 
 main {
   max-width: 920px;
+  margin: auto;
 }
 
 header {


leo:
Thanks for the effort you've put into this PR, but please make sure to discuss it with the maintainers the next time. We'd like to keep the current design! 😊

PR #322, opened 2018-02-06
Add some information about ignoring option in help

mmmm1998:
If you want ignore file `foo.txt`, you should use command `serve --ignore foo.txt`.
If you want ignore **directory** foo, you should use command `serve --ignore foo/`.
It's obviously, so add information about ignoring directory in help.


mmmm1998:
Well, I don't know, how I can change options.js.snap. Any suggestion?

fmiras:
@mmmm1998 you should make use on cli
```bash
ava --update-snapshots
```

mmmm1998:
@fmiras Thanks!

leo:
Thank you for the effort you've put into this!

But the `--ignore` flag is now gone in favor of using `rewrites` to rewrite your path to a 404 route.

This is now possible by supplying a `serve.json` file. All thanks to `serve-handler`, as described in the notes for the [new release](https://github.com/zeit/serve/releases/tag/7.0.0).

PR #323, opened 2018-02-07
Download directory as zip feature

mmmm1998:
Now user could download an open folder entirely as zip archive, which is comfortably, when you share a lot of files. 
default
I could not create nice button for downloading, which really fit for view page, but I'm ready to fix it, if anyone tell me how create beautiful icon or something else.


leo:
Thank you for the effort you've put into his! 😊 

But please consider creating [an issue](https://github.com/zeit/serve/issues) the next time before you start tackling the topic, so that we can discuss it and agree on it. Because in this case, I don't think it's a good idea to merge this into the project, since we're trying to keep the core as small as possible.

🙏

mmmm1998:
It is a pity, because many file sharing server does not have this feature, which is quite inconvenient sometimes.
But as you wish.

PR #326, opened 2018-02-13
Temporary work around `mime-types` limitation

philipp-spiess:
This will temporarily add a mime type for `.wasm` files as per suggestion in https://github.com/jshttp/mime-types/issues/44#issuecomment-365326416. As soon as an update to the upstream library is made, this change can be removed again.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -16,6 +16,12 @@ const isPathInside = require('path-is-inside')
 // Utilities
 const renderDirectory = require('./render')
 
+// Register the mime type for `.wasm` files. This change is currently not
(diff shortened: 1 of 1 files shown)


philipp-spiess:
This fix won't work since files are streamed with [`send`](https://github.com/pillarjs/send) which apparently depends not on `mime-types` but on (`node-mime`](https://github.com/broofa/node-mime).

`.wasm` support was added there in https://github.com/broofa/node-mime/commit/3f775bafa8bff0c1a4bd94bb07ab3065676e44f6 so maybe we can simply upgrade that module?

Issue #325, opened 2018-02-13
No way to serve on localhost only


xrisk:

paulogdm:
This doesnt make any sense at all. Can you please explain?

xrisk:
If I type `serve`, then my files get served to anybody on my network, whereas I want it to be only served on localhost. 

There is no way to do this judging from the output from `serve --help`. Ideally, `serve` would accept an address to bind on, instead of simply binding to `0.0.0.0`.

leo:
Great catch, we should definitely have a flag for that! 👍 

```
serve --local
```

or 

```
serve --only-local
```

(I think the first one is better).

And then it only listens on `localhost` instead of `0.0.0.0`.

xrisk:
It might be better to accept an address to bind on, instead of arbitrarily constraining the user between a choice of `0.0.0.0` / `localhost`. 

A particular machine could have more than 1 IP address for instance. You should be able to serve on once of those IPs without affecting the other.

leo:
Closed in https://github.com/zeit/serve/pull/327#pullrequestreview-99129079

PR #327, opened 2018-02-14
Added support for a `--local` flag

fmiras:
According to #325. This creates de localhost flag to serve content without sharing it to the private network.

Usage:
```bash
serve --local
```


leo:
What a wonderful PR, thank you! 😍

PR #328, opened 2018-02-14
Fixes wrong mime type for `.wasm` files

### LINKED ISSUE #303: Wrong mime-type for .wasm
Mime-type for `.wasm` should be `application/json`, but currently its `text/html`. It can be fixed by upgrading [mime-db](https://www.npmjs.com/package/mime-db) which is dependency of [mime-types](https://www.npmjs.com/package/mime-types). The current version (accroding to the lock file) is `1.30.0`

giuseppeg:
Fixes #303 

@leo this is a temporary fix. If you don't want to merge I can use a fork.


giuseppeg:
discarding in favor of #331

PR #338, opened 2018-02-27
Relaxes version constraints for dependencies

lourd:
Now that y'all are using yarn and have a `yarn.lock` file, you don't need to specify exact versions for your dependencies. Relaxing the constraints will allow for compromise and optimizations regarding shared dependencies when installing all of the dependencies for an application.


leo:
Thanks for the effort, but I don't think lock files prevent patches from getting installed (which is what `^` is doing)? We can't have that because a lot of people don't play by the rules and release breaking updates as patches...

lourd:
To use the lock file to prevent different versions from being installed, including patches, use the [`--frozen-lockfile`](https://yarnpkg.com/en/docs/cli/install#toc-yarn-install-frozen-lockfile) option for `yarn install`.

timneutkens:
Lockfiles are not respected by NPM/Yarn while installing, only when you clone serve and run `yarn`. When you `yarn add serve` or `yarn global add serve` it'll install the *latest minor* version of all dependencies, which is not what we want, since breaking patches/minors are often released and would break `serve`. This is the reason we do locking in the package.json.

leo:
Exactly! Closing this, now that I got a second opinion on it.

But again: Thanks for your effort, @lourd! 😊

PR #339, opened 2018-02-28
Ensure `0` cache value triggers no cache

EdwardSalter:
For some reason, setting a cache value of 0 in both the CLI and the API returns the value as a string.
The `args` package must not treat 0 as a number for some reason.

Also modified Cache-Control header for preventing caching to be more aggressive by setting the no-store and must-revalidate options as Chrome was still accessing assets using the cache.


leo:
Thank you!

PR #340, opened 2018-03-03
Add additional entry point option

gebeto:
Added new option that allows to change `index.html` entry point to something else, and if your new entry point file is not exists thats use `index.html`.


lib/server.js
```diff
@@ -135,10 +135,15 @@ module.exports = coroutine(function*(req, res, flags, current, ignoredFiles) {
       res.end()
       return
     }
-
-    let indexPath = path.join(related, '/index.html')
+    
+    const { entry } = flags
+    let indexPath = path.join(related, '/' + entry)
     res.setHeader('Content-Type', mime.contentType(path.extname(indexPath)))
```
stripedpajamas: Shouldn't this `setHeader` come after `indexPath` is definitively determined? If an invalid/non-existent entry file is specified, the logic is that it will default to index.html, but `setHeader` will have already been called.

gebeto: Yep, but in current code without my feature, exists the same problem, `setHeader` called only once and if `index.html` will not exists, that will error too, there are no check for not existing `index.html`

 I am just want to add the feature and use it, bug fixing will be a next steps😄

stripedpajamas: good call 👍

gebeto: sorry if I'm wrong😅

PR #352, opened 2018-04-11
Only check for update in non-production

### LINKED ISSUE #351: New update-checker breaks compatibility with Read Only filesystems
The Pull Request: https://github.com/zeit/serve/pull/347 has introduced a new update checking module. This module tries to create a temporary folder and file under `/tmp` - `/tmp/update-check/serve-latest.json`.

However in our deployments, as our file-systems are Read Only the new update breaks. The previous point release `6.5.3` was compatible and working with our setup, whereas `6.5.4` has broken this compatibility.

Operating System: Alpine Linux
Node: 8.11.1

A stack trace can be found below: 
```
(node:6) UnhandledPromiseRejectionWarning: Error: EROFS: read-only file system, mkdir '/tmp/update-check'
(node:6) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:6) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:6) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open '/tmp/update-check/serve-latest.json'
(node:6) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
```

thekemkid:
Should fix #351

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -82,20 +82,23 @@ let { port } = flags
 
 detect(port).then(async open => {
   const { NODE_ENV } = process.env
-  const update = await checkForUpdate(pkg)
 
-  if (NODE_ENV !== 'production' && update) {
-    const message = `${bold(
-      'UPDATE AVAILABLE:'
-    )} The latest version of \`serve\` is ${update.latest}`
+  if (NODE_ENV !== 'production') {
+    const update = await checkForUpdate(pkg)
 
-    console.log(
-      boxen(message, {
-        padding: 1,
-        borderColor: 'green',
-        margin: 1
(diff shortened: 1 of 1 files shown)


leo:
Thank you! 😊 Released with the latest update.

PR #377, opened 2018-05-28
Removed useless dependencies

leo:
As suggested by @rauchg, we were able to drop 3 more dependencies! 🚀 

Now we're down to just 6 (4 of them being ours).

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -1,21 +1,23 @@
 #!/usr/bin/env node
 
 // Native
+const http = require('http');
 const path = require('path');
+const fs = require('fs');
+const {promisify} = require('util');
 
 // Packages
 const Ajv = require('ajv');
-const dotProp = require('dot-prop');
 const checkForUpdate = require('update-check');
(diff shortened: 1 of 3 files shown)


bin/serve.js
```diff
@@ -154,29 +153,58 @@ const startEndpoint = (endpoint, config) => {
...
 	const config = {};
 
-	for (const file of Object.keys(paths)) {
+	for (const file of files) {
 		const location = path.join(entry, file);
-		const prop = paths[file];
+		let content = null;
+
+		try {
+			content = await readFile(location, 'utf8');
```
pranaygp: `await` in a loop seems bad?

leo: Talked this out via DM. Usually a great idea, but we want to stop the loop immediatley once we notice that a file is readable.

But at the same time allow `readFile` to error in order to cover files that don't exist (which shouldn't make us loose the rest of the values, which is what `Promise.all` would do, e.g).

Issue #380, opened 2018-05-29
ERROR: URL is not defined


gdereese:
When running the following command on v7.0.1 of the package:
`serve --listen tcp://127.0.0.1:1234`
the following error is logged to the console and the program terminates:
`ERROR: URL is not defined`

This is in context of a package running in Node 8.11.2.

jaeseok-park:
I have the same issue.

jaeseok-park:
Temporarily, I added a line in bin/serve.js.

```
#!/usr/bin/env node

// Native
const http = require('http');
const path = require('path');
const fs = require('fs');
const {promisify} = require('util');
const {URL} = require('url');             // This is the added line.
```

leo:
Fixed in https://github.com/zeit/serve/pull/381

PR #381, opened 2018-05-29
Fix "URL is not defined" error

jaeseok-park:
I have made a PR for the issue #380.
This is the first PR for public github open source.
If I made a mistake, please advise me.
Thank you.

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -5,6 +5,7 @@ const http = require('http');
 const path = require('path');
 const fs = require('fs');
 const {promisify} = require('util');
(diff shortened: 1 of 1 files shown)


leo:
It seems like you're using an older version of Node.js (the `URL` class is now part of the `global` scope and therefore available without importing the module).

But supporting that Node.js version of yours? I mean, why not! Thank you!

jaeseok-park:
My Node.js version is v8.11.2 LTS FYI.
Thank you for your approval.

PR #407, opened 2018-06-07
Show server listen address properly

iczero:
When a listen address is specified as `-l tcp://addr:port`, the listen address is always shown as `localhost`, instead of what is specified in `addr`.


leo:
Amazing, thank you!

bodaz:
the info log appears to be this now after v8.1.1:

screen shot 2018-06-08 at 09 35 08

iczero:
That is normal, as the listen address is [::] (the IPv6 equivalent of 0.0.0.0)

leo:
@bodaz https://github.com/zeit/serve/pull/410

PR #410, opened 2018-06-08
Correctly render hostname

leo:
🤦‍♂️ 😄 

This was introduced in https://github.com/zeit/serve/pull/407.

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -156,7 +156,8 @@ const startEndpoint = (endpoint, config) => {
 		if (typeof details === 'string') {
 			console.log(info(`Accepting connections at ${details}`));
 		} else if (typeof details === 'object' && details.port) {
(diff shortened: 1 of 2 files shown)


bin/serve.js
```diff
@@ -156,7 +156,8 @@ const startEndpoint = (endpoint, config) => {
 		if (typeof details === 'string') {
 			console.log(info(`Accepting connections at ${details}`));
 		} else if (typeof details === 'object' && details.port) {
-			console.log(info(`Accepting connections at http://${details.address}:${details.port}`));
+			const address = details.address === '::' ? 'localhost' : details.address;
```
colinbate: I don't think that `::` is the same as `localhost`. I believe `::1` is localhost, and `::` is similar to `0.0.0.0` in IPv4.

PR #418, opened 2018-06-12
Add `--config` for custom paths to `serve.json`

tohjustin:
Add `--config` for users to specify custom paths to their `serve.json` file. 

Below is an example of configuring `serve` to use `serve.json` residing in the `scripts` directory:
```
serve --config scripts/serve.json
```

This closes zeit/serve#401.


leo:
This is amazing. Thank you!

PR #434, opened 2018-06-18
Read port from environment variable `PORT`

### LINKED ISSUE #433: Reading port from env
Since #374, the port number is no longer read from `process.env.PORT` (then fallback to `5000`).

Any chance to bring it back? I guess should modify [`serve.js:133`](https://github.com/zeit/serve/blob/master/bin/serve.js#L133).

compulim:


leo:
Thanks!

PR #439, opened 2018-06-24
Handle DNS lookup failures

### LINKED ISSUE #437: UnhandledPromiseRejectionWarning: Error: getaddrinfo ENOTFOUND <my hostname>
~~~bash
src% mkdir tmp
src% cd tmp
tmp% echo "Hello, world!" > index.html
tmp% serve
(node:39744) UnhandledPromiseRejectionWarning: Error: getaddrinfo ENOTFOUND flower
    at errnoException (dns.js:50:10)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:92:26)
(node:39744) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:39744) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
^C
INFO: Gracefully shutting down. Please wait...
tmp% hostname
flower
tmp% nslookup `hostname`
Server:		10.245.22.6
Address:	10.245.22.6#53

** server can't find flower: NXDOMAIN
tmp% serve --version
9.1.0
~~~

just-boris:
When lookup fails, it throws an exception, that prevent other code from being executed

fixes #437

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -173,10 +173,14 @@ const startEndpoint = (endpoint, config, args) => {
 			localAddress = details;
 		} else if (typeof details === 'object' && details.port) {
 			const address = details.address === '::' ? 'localhost' : details.address;
-			const {address: ip} = await lookup(os.hostname());
 
 			localAddress = `http://${address}:${details.port}`;
-			networkAddress = `http://${ip}:${details.port}`;
+			try {
+				const {address: ip} = await lookup(os.hostname());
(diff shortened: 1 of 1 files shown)


leo:
Amazing. Thank you!

Issue #423, opened 2018-06-13
check for update: skip for production deployments, implement timeout


patrikbeno:
all of this looks OK before #374 was introduced.

Timeout was introduced in #360 and update check was skipped entirely in production.

Now it's all back to naive and potentially breaking implementation.

serve@8.2.0

leo:
The update check is happening behind a [catch](https://github.com/zeit/serve/blob/master/bin/serve.js#L32-L39) statement. It's not possible for it to crash the program.

patrikbeno:
Following is possible:
- hang during startup while waiting for connection
- print irrelevant warnings during startup, e.g. SSL failure
- delay startup while performing irrelevant update check

I consider this a bug, because I cannot prevent any of this.

At least consider a `--no-update-check` option

leo:
There is a [timeout](https://github.com/zeit/update-check/pull/9) in place, so that cannot happen

patrikbeno:
@leo, it is irrevant if you are not interested in any kind of update check. In production, for example.
All I ask for is the option to disable the update check. I am interested in `serve` feature, not `update`, especially in production environment.

Of course, update check is useful, I don't dispute that, especially during development, but in production deployments, all it does is useless and out of context.

patrikbeno:
@davemorrissey, in my case, I run it in docker, and I was able to short-circuit the whole process by following:

```
# dummy registry effectivelly disables update check and avoids irrelevant connection errors and startup delays
RUN echo "registry=http://localhost" > .npmrc
ENV NODE_ENV="production"
# /etc/hosts entry avoids DNS lookup delays and errors when resolving self hostname
CMD printf "127.0.0.1 $(hostname)" >> /etc/hosts && serve -l ${PORT:-80} -s
```

Issue #369, opened 2018-05-20
Why does serve silently check for updates?


Winona-Ryder:
It silently checks for updates in the background and I can't see this documented anywhere.

1. Why does it need to check for updates? this activity appears as a random internet connection where none was expected and gives off a spyware vibe (I know it's not, but that's the vibe it gives off).

2. Why isn't this documented anywhere (that I can find)?

3. How do I disable it?

Note: I didn't install the package manually, it came as dependency of other packages.

IMO, it should not be checking for updates by default, or at least not when the package is obviously not installed globally (which is a bad practice IMO).

leo:
Sorry for the inconvience we've caused, but plenty of the command line interfaces installed on your device through npm are doing that.

We're simply using [update-check](https://github.com/zeit/update-check) to check for updates of the package, in order to keep it updated.

Among all of those is `npm` itself, by the way.

Both if the package is installed locally and globally, it makes sense to check for updates in order to let the user know if there's a new one, so that it's always up-to-date in case there was an important update released.

Currently, there is no way to disable this as we haven't received any other complaints yet, but I think it might actually be a great idea to provide some kind of environment flag for that. 👍 

Closing this, however, as that would be something for [update-check](https://github.com/zeit/update-check) itself.

Thank you for your feedback!

leeyeh:
Here is another complaint. 😂 

Notifications about a newer version helps, however, this is not always the case when the package is used as a dependency.

For the users of my project, they do not know which package I'm using or how I use it to serve files. The update notification does not make  too much sense for them.

As the project author, I prefer to stick to a working version (using a package-lock.json to keep everything consistent) and use some dependency monitor services (i.e snyk.io) to keep myself notified with some important security updates.

`serve` is not a CLI for a service (like `npm`), a working version is good enough for me.

PR #457, opened 2018-07-27
Added `NO_UPDATE_CHECK` environment flag

leeyeh:
#369 #423 

Usage:

```sh
NO_UPDATE_CHECK=1 serve
```


leo:
Can we only use the env variable for that? I wouldn't add another option.

leeyeh:
@leo Option removed.

leo:
Thank you!

PR #472, opened 2018-09-07
Disable clean URLs

CalRobert:
Disable serve-handler's clean url's by default


leo:
Thank you for this PR, but we're not considering changing the defaults (major breakage).

PR #487, opened 2018-11-12
Added support for compression

leo:
This ensures the responses are compressed by default.

It can be disabled using `-u` or `--no-compression`.


gaearon:
Thank you!

Issue #498, opened 2018-12-14
Issues setting config file (`serve.json`)


arel:
Just started using `serve` and it's great! But, I had a hard time figuring out how to configure it with a custom `serve.json` file. There were several contributing factors:

* Absolute paths to `serve.json` are not found.
    ```bash
    serve --config /app/serve.json build  # SILENTLY FAILS
    ```

    * This seems like a bug
    * At a minimum, this should be documented in the README, if it's intended behavior

* Config path is relative to the path being _served_ not the current working directory.
    * I would have expected the path relative to where the command was called.
    * At a minimum, this should be documented clearly in the README

* `serve` fails silently when `--config` is set but file is not found.
    * I would expect an error if the file is missing.
    * At minimum I would expect a warning (similar to when config file is found: `INFO Discovered configuration in ../serve.json`)

* Documentation does not clearly indicate how to set config file, or what a `serve.json` file looks like. An example would be helpful here.

FlorianBurgevin:
+1

jespertheend:
I was having a hard time figuring out how serve.json works as well. I didn't know the path was relative to the directory being served. It's working now. Thanks @arel!

bensleveritt:
Being relative to the served directory really threw me too. Would be good to either have that documented, or just relative to the CWD.

warren-bank:
simple fix.. please [patch](https://github.com/zeit/serve/tree/11.0.2/bin/serve.js#L249)

__current code__:

```javascript
const loadConfig = async (cwd, entry, args) => {
	const files = [
		'serve.json',
		'now.json',
		'package.json'
	];

	if (args['--config']) {
		files.unshift(args['--config']);
	}

	const config = {};

	for (const file of files) {
		const location = path.join(entry, file);
		//...
	}
	//...
}
```

__better methodology__:

```javascript
const loadConfig = async (cwd, entry, args) => {
	const files = [
		path.join(entry, 'serve.json'),
		path.join(entry, 'now.json'),
		path.join(entry, 'package.json')
	];

	if (args['--config']) {
		files.unshift(
			path.isAbsolute(args['--config'])
				? path.resolve(args['--config'])
				: path.join(entry, args['--config'])
		);
	}

	const config = {};

	for (const file of files) {
		// [line removed] const location = path.join(entry, file);
		//...
	}
	//...
}
```

PR #502, opened 2019-01-19
Always throw error if config file is specified

gamgi:
As mentioned in #498 current behaviour is to silently fail if `--config` file is not found. This can result in running the server with an unintended configuration without the user knowing about it.

The change in this PR ensures that errors are thrown if the `--config` file is explicitly specified.

Example behaviour:
```
$ serve -c nonexistent.json
ERROR: Not able to read /foo/nonexistent.json: ENOENT: no such file or directory, open '/foo/nonexistent.json'
```

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -264,7 +264,7 @@ const loadConfig = async (cwd, entry, args) => {
 		try {
 			content = await readFile(location, 'utf8');
 		} catch (err) {
-			if (err.code === 'ENOENT') {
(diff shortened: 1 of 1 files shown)


jakearchibald:
Ugh, yes please, merge this. This caught me out, especially as the config path is resolved against the content path, which is pretty unusual.

longzheng:
I've also been caught out by this issue leading to a very confusing/frustrating situation, it is extremely subtle that the config was/was not being loaded.

leerob (approved):

PR #504, opened 2019-02-06
Update help text.

GonzaloRizzo:
By default, in practice and looking at the code, it seems to only be listening in localhost so we should update the help text so we don't confuse users.

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -61,7 +61,7 @@ const getHelp = () => chalk`
       {bold $} {cyan serve} --version
       {bold $} {cyan serve} [-l {underline listen_uri} [-l ...]] [{underline directory}]
 
-      By default, {cyan serve} will listen on {bold 0.0.0.0:5000} and serve the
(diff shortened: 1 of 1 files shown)


r0mflip:
🤔 I don't think that helps in anyway, it majorly disproves whats happening under.

GonzaloRizzo:
What do you mean? Isn't serving in `localhost:5000` the default behaviour?

r0mflip:
Serving on `localhost:5000` is different from `0.0.0.0:5000`.

`localhost` listens only on the loopback interface(logical), `0.0.0.0` listens on all network interfaces(IPv4 - physical and logical). It's through what interface the server can be contacted from the outside world(or from within the host).

Node listens on `0.0.0.0` or `::` [by default](https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback).

leo:
Closing. The above comment is correct.

PR #523, opened 2019-04-13
Missing option --no-compression in help message

mbalabash:
Hi
I'm couldn't find any info on how to disable compression in the --help message.
But this feature actually exists. I found it in the source code.

These changes are adding a missed --no-compression option in the help message.


mbalabash:
Hello, @leo 
Could you check this?

PR #541, opened 2019-06-28
Add ssi #include support

SylRob:
SSI is an old techno but,
adding SSI include support to this package would be really helpful for the devs that still (have to) use it.

I added the _--ssi_ : String key,
if found, and in is correct URL format :

- the serve-handler function param "_createReadStream_" is extended
- the html file is read(_readFileSync_) and transformed into a string,
- the SSI function is trigger with the --ssi URL and the file as a string as parameter
- the SSI function look for < !--#include virtual="/path/to/the/index.html" -- > patterns
- if match is found, dl the file **synchronously**  with the package _sync-request_ and the _--ssi_ URL as base and replace the match with the dl file content
- finally return the formatted html, this html is pushed in a _new Readable()_ and return in _createReadStream_

Note: the **synchronously** in point 5 will cause time out if the file does not exist on the server...
this could be improve / prevent

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -4,6 +4,8 @@
 const http = require('http');
 const path = require('path');
 const fs = require('fs');
+const SSI = require('./ssi.js');
+const Readable = require('stream').Readable;
 const {promisify} = require('util');
 const {parse} = require('url');
 const os = require('os');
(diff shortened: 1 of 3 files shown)


leerob:
We don't plan on including support for this right now 🙏 Thank you regardless.

PR #544, opened 2019-07-04
Updated README for better understanding where to place the serve.json

philippescher:
It took me more than 20 minutes to understand where to put the serve.json. So I updated your README.


leo:
Thank you!

PR #572, opened 2020-01-06
Fix undefined local network address

wabscale:
In the event that serve is started when the host machine is not connected to the internet, the local network address will show up as undefined. Adding a simple check fixes this. When this is the case, the local network address won't appear anymore.

- Fix undefined local network address

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -225,7 +225,7 @@ const startEndpoint = (endpoint, config, args, previous) => {
 			const ip = getNetworkAddress();
 
 			localAddress = `${httpMode}://${address}:${details.port}`;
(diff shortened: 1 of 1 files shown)


leerob:
`null` is not an improvement over `undefined` here, so I'm going to close this as won't fix. Thank you regardless 🙏

wabscale:
If you read a few lines down, if networkAddress is `null` then it does not print it out. It does not print `null`.

wabscale:
Here is how it looks right now in the master branch:

before

And here is what it looks like after my change:

after

leerob:
My apologies - I definitely read this wrong!

leerob (approved):
Thank you!

PR #579, opened 2020-02-25
Add flag for disabling port switching

jespertheend:
https://github.com/zeit/serve/issues/365

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -202,7 +202,7 @@ const startEndpoint = (endpoint, config, args, previous) => {
 		: http.createServer(serverHandler);
 
 	server.on('error', (err) => {
-		if (err.code === 'EADDRINUSE' && endpoint.length === 1 && !isNaN(endpoint[0])) {
+		if (err.code === 'EADDRINUSE' && endpoint.length === 1 && !isNaN(endpoint[0]) && args['--no-port-switching'] !== true) {
 			startEndpoint([0], config, args, endpoint[0]);
 			return;
 		}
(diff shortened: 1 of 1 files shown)


bin/serve.js
```diff
@@ -370,6 +370,7 @@ const loadConfig = async (cwd, entry, args) => {
 			'--no-etag': Boolean,
 			'--symlinks': Boolean,
 			'--cors': Boolean,
+			'--no-port-switching': Boolean,
```
leerob, line 373: Could you add documentation for this?

jespertheend: I've added it to --help. Would you like to have it added somewhere else as well?

leerob: True, that will probably suffice. Thank you.

leerob (approved):

JamesHenry:
@leerob Sorry to be commented on a merged PR, but it seems a shame to open an issue just for this:

It looks like there hasn't been a release of the library for a year now - please could you cut one so that we can easily make use of this change via NPM?

Thanks!

JamesHenry:
Hmm additionally @jespertheend was it your intention that this _doesn't_ exit with code 1?

Having installed it via the git commit and run it, it always seems to hit `INFO: Gracefully shutting down. Please wait...` and exits with code 0 which is unintuitive I think

```
ERROR: Failed to serve: Error: listen EADDRINUSE: address already in use :::4400
    at Server.setupListenHandle [as _listen2] (net.js:1318:16)
    at listenInCluster (net.js:1366:12)
    at Server.listen (net.js:1452:7)
    at startEndpoint (/Users/james/.../node_modules/serve/bin/serve.js:220:19)
    at /Users/james/.../node_modules/serve/bin/serve.js:445:3

INFO: Gracefully shutting down. Please wait...
```

jespertheend:
Ah interesting, I hadn’t even thought about that to be honest. Non-zero seems more intuitive yeah. I was just reusing the shutting down code that was already in place I believe. So it should behave the same as when an error other than EADDRINUSE is thrown.

leerob:
This has now been released 👍

JamesHenry:
Thanks @leerob, just installed it instead of the commit reference and everything looks good!

PR #599, opened 2020-05-26
Document CORS option

brandonarbini:
The option to enable CORS has returned, but the documentation did not. The spacing may be off since I did this directly in GitHub.


TooTallNate:
Thanks!

PR #607, opened 2020-06-11
add undocumented option flag to the help command output

Oscar-21:
The -p port command is not available in the help command. I assumed it wasn't available as a result of this. I think documenting tihs in the help command output is a good idea.


dgurns:
Looks like the `-l` command accomplishes this, for example `serve -l 4000` will serve your site on port 4000.

However I also ran into this initial confusion and it would be great to document some of these flags in the main README.

leerob (approved):
Thank you!

Issue #438, opened 2018-06-21
Allow removal of a header


lorensr:
I'd like to be able to set a default header and then override it to nonexistance. When I do the below, a header with null shows up, versus the header disappearing. Ditto with `max-age=0`. And when I do `"headers": []`, it gets 86400 from the first rule.

![image](https://user-images.githubusercontent.com/251288/41744694-9803be1a-7572-11e8-9fb9-ec87dd70a42f.png)

![image](https://user-images.githubusercontent.com/251288/41744702-9b6b944c-7572-11e8-9b0e-6d3c6e195020.png)

cc @leo

leo:
Setting a header to `null` should definitely remove it! 👍

WellingGuzman:
Hey @leo, I tried to use `null` but still serves it as `null` instead of deleting it.

I made a PR that solves it.

Ref: https://github.com/zeit/serve-handler/pull/45

WellingGuzman:
This ticket can be closed. It was solved by https://github.com/zeit/serve-handler/pull/45

sidvishnoi:
As per vercel/serve-handler#45, it should've been fixed in latest version of serve (11.3.2, which uses serve-handler 6.1.3, which includes mentioned PR). But I'm getting the following error
```
$ serve
INFO: Discovered configuration in `serve.json`
ERROR: The configuration you provided is wrong:
should be string
{"type":"string"}
```

<details>
<summary>serve.json</summary>

``` json
{
	"headers": [
		{
			"source": "**/*.js.br",
			"headers": [
				{ "key": "Content-Encoding", "value": "br" },
				{ "key": "Content-Type", "value": "text/javascript" },
				{ "key": "Content-Disposition", "value": null }
			]
		}
	]
}
```
</details>

wellingguzman:
@sidvishnoi I noticed they implemented a schema validation which doesn't allow `null`. Hopefully https://github.com/vercel/schemas/pull/58 and https://github.com/vercel/serve/pull/629 could solve this problem.

It requires to upgrade `ajv` and update the instance to allow null values.

juyuwol:
I'm also having trouble with not being able to remove the HTTP Header.

Currently, because Vercel automatically sets the Content-Disposition header, so I couldn't specify the file name as the download attribute of the <a> tag.

sunlee-newyork:
hi, using `null` is returning a deployment fail error `'headers[0].headers[1].value' should be string`

[...]

PR #629, opened 2020-09-01
Allow null on headers value

wellingguzman:
Solves: vercel/serve#438 and must be merged together with https://github.com/vercel/schemas/pull/58


leerob:
We are not currently planning to support null headers. Thank you for your PR regardless 🙏

PR #634, opened 2020-10-23
Added logging

danhab99:
Added an option to output an access log including time, url, and remote ip.


leerob:
We are rewriting the project in TypeScript - if you are still interested in contributing, could you open a PR against this branch? https://github.com/vercel/serve/pull/706

Thank you!

PR #647, opened 2021-03-20
add --sab to enable SharedArrayBuffer

tlhunter:
This is a super-specific change that I'm not sure makes a lot of sense being merged into serve. Basically, there's now an `--sab` flag to enable [two headers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements) required to support `SharedArrayBuffer` instances in Firefox (Chrome seems to support the feature without the headers).

```sh
$ serve . --sab
```

The approach isn't necessarily the most ideal as I can imagine other headers being needed to support other web technologies. I personally want this flag so that readers of my upcoming book can spin up a local server with a one-liner and experiment with `SharedArrayBuffer`.

Perhaps a more general-purpose solution that makes sense for serve is to accept a list of headers provided via flags to be returned with each request, like so:

```sh
$ serve . --header Cross-Origin-Opener-Policy=same-origin --header Cross-Origin-Embedder-Policy=require-corp
```

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -87,6 +87,9 @@ const getHelp = () => chalk`
 
       -C, --cors                          Enable CORS, sets \`Access-Control-Allow-Origin\` to \`*\`
 
+      --sab                               Enable SharedArrayBuffer, sets \`Cross-Origin-Opener-Policy\` to
+                                          \`same-origin\` and \`Cross-Origin-Embedder-Policy\` to \`require-corp\`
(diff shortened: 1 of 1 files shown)


leerob:
Would you mind opening this as an issue first? The headers approach seems to make more sense to me. Then we can see if anyone else chimes in needing it! 🙏

PR #659, opened 2021-07-07
Fix incorrect empty checking: networkAddress will never shown

anjianshi:


leerob:
https://github.com/vercel/serve/pull/661

PR #661, opened 2021-07-16
Fix network not being shown anymore

### LINKED ISSUE #658: Why did you remove message?
Why did you remove "- On Your Network:" ip in log output in 12.0.0? This was very helpful!

marcofugaro:
Close #658

I am astonished that a PR like #572 made it into production.

`networkAddress` is always null, for this reason the network address is not being shown anymore on any machine. I believe `ip` is what the author of that PR meant.


leerob (approved):
Thank you!

PR #678, opened 2021-10-27
add nextjs server start

an-dr-eas-k:
Added nextjs server start with --nextjs option. For simple projects no need to code custom server.js file e.g. to easily get encryption.


leerob:
We'd recommend using Next.js for this directly 👍

Issue #672, opened 2021-09-21
macOS Monterey occupies port 5000, prevents `npx serve` using default port


peterfriese:
It seems like the AirPlay Receiver process on macOS Monterrey uses ports 5000 (which seems to be the default for `serve`) and 7000:

https://developer.apple.com/forums/thread/682332

kachkaev:
The new default port is 3000, so I guess that the issue can be closed 👍

karlhorky:
@peterfriese Seems like this can be closed, right?

PR #680, opened 2021-10-29
Change default port from `5000` to `3000`

kud:
Due to a conflict with AirPlay, it could be interesting to change the default port: https://developer.apple.com/forums/thread/682332

related: https://github.com/vercel/serve/issues/672
info: https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -63,7 +63,7 @@ const getHelp = () => chalk`
       {bold $} {cyan serve} folder_name
       {bold $} {cyan serve} [-l {underline listen_uri} [-l ...]] [{underline directory}]
 
-      By default, {cyan serve} will listen on {bold 0.0.0.0:5000} and serve the
+      By default, {cyan serve} will listen on {bold 0.0.0.0:3000} and serve the
       current working directory on that address.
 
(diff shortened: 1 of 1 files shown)


kachkaev:
3000 is used by Next.js and CRA by default, which means that ports will often clash in practice. Having said that, I don’t know what port to suggest.

Nice-looking numbers like `6000`, `9000` or `10000` are all used for something else, [according to Wikipedia](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers). Maybe 3030? 🤷

leo:
I don't think it will be common that people will run tools like Next.js and CRA and the same time as `serve`, but if they do, `--listen <port>` can be used for defining a custom one.

`3000` as the new default sounds good to me!

leo:
Thank you!

kud:
My pleasure 😊

mariojfduarte:
If you have a Docker Swarm mode cluster set up with traefik as a load balance to do wtv, don't forget to change alll your docker files that use _serve_ default port in order to prevent **Bad Gateway** connection errors.

`
...
RUN yarn global add serve 
...
CMD serve -s **-p 5000** build
...

`

PR #681, opened 2021-11-04
Switch to npm

leo:
🚢


barelyhuman:
maybe add a `.yarnrc` with 
```
--install.no-lockfile true
--add.no-lockfile true
```

to avoid someone accidentally adding `yarn.lock` during contribution and missing it in code reviews

GrosSacASac:
https://nodejs.org/api/packages.html#packagemanager

PR #700, opened 2022-05-21
Fix not showing network address on Node 18

junkycoder:
Since Node 18, `family` is a number 👉 https://nodejs.org/api/os.html#os_os_networkinterfaces


gamemaker1:
This was reverted in Node `18.4.0`.

leerob:
Given this was reverted, I think we can close this 🙏

PR #580, opened 2020-02-26
Add support for PFX or PKCS12 encoded certificates

schemburkar:
Added support for PFX or PKCS12 encoded certificates by providing additional command line parameters

e.g. 
```
$ serve -l 5000 --ssl-cert "cert.pfx" --ssl-pass "password.txt" --ssl-format "pfx"
```
New Command line options

**--ssl-format**:  Optional format of the SSL/TLS certificate. Supported formats: pem (default) and pfx

When ```--ssl-format``` is omitted, the default format (pem) will be used.

Edit: ```--ssl-pass``` is supported for both pem and pfx formats.
Edit: Removed ```--ssl-passphrase``` in favor of existing ```--ssl-pass``` option

PR #706, opened 2022-07-07
refc: rewrite the project in typescript

### LINKED ISSUE #705: Add ESLint and move to TypeScript
Copy setup from https://github.com/vercel/ms

gamemaker1:
## Related Issues

- Closes #705

## Description

This PR:

- rewrites the library in Typescript
- makes this package an ES module.
- adds ESLint and Prettier
- adds a contributing guide
- adds `lint` and `test` steps to the GitHub workflow
- adds issue and pull request templates

## Checklist

- [x] The issues that this PR fixes/closes have been mentioned above.
- [x] What this PR adds/changes/removes has been explained.
- [x] All tests (`pnpm test`) pass.
- [x] The linter (`pnpm lint`) does not throw an errors.
- [x] All added/modified code has been commented, and
      methods/classes/constants/types have been annotated with TSDoc comments.


gamemaker1:
I've tried to stick as close to the style guide as possible, but please let me know if there's something I missed and should change!

schemburkar:
@gamemaker1 I have ported my work in #580 to code based on your current PR. How do you want me to push this? Directly to  this PR or separate one?
cc: @leerob

gamemaker1:
@schemburkar That was really fast 🎉 Thanks for the amazing work on the PR!

I think it makes sense to keep the certificate PR separate and merge it in after this PR.

I want to make a couple of other PRs too, like the logging one and changes from useful forks like https://github.com/warren-bank/node-serve, but I'll wait until this one is merged and then make those separately. It'll make it easier for the Vercel team to review changes :]

config/husky/pre-commit
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+# config/husky/pre-commit
+# Run `lint-staged` before every commit.
+
+. "$(dirname "$0")/_/husky.sh"
+
+FORCE_COLOR=2 pnpm lint-staged
```
leerob, line 8: What does `FORCE_COLOR` do?

source/main.ts
```diff
@@ -0,0 +1,174 @@
...
+
+// Default to listening on port 3000.
+// Disabling as this is not an unnecessary check.
+// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+if (!args['--listen'])
+  args['--listen'] = [
+    [process.env.PORT ? parseInt(process.env.PORT, 10) : 3000],
+  ];
+// Ensure that the user has passed only one directory to serve.
+if (args._.length > 1) {
```
leerob, line 75: We should make a future issue to start adding tests with `vitest` 😄

source/types.ts
```diff
@@ -0,0 +1,89 @@
...
+export declare interface Options {
+  '--help': boolean;
+  '--version': boolean;
+  '--listen': ParsedEndpoint[];
+  '--debug': boolean;
+  '--single': boolean;
+  '--config': Path;
+  '--cors': boolean;
+  '--no-clipboard': boolean;
+  '--no-compression': boolean;
```
leerob, line 77: Should we sort this alphabetically so it matches the list below in `cli.ts`?

source/utilities/config.ts
```diff
@@ -0,0 +1,126 @@
...
+ * @param args - The arguments passed to the CLI.
+ *
+ * @returns The parsed configuration.
+ */
+export const loadConfiguration = async (
+  cwd: string,
+  entry: string,
+  args: Options,
+): Promise<Configuration> => {
+  const files = ['serve.json', 'now.json', 'package.json'];
```
leerob, line 30: There was an error thrown in the previous file is you use `now.json` since it's deprecated - I wonder if we rip that logic out?

leerob: e.g. if we wrote a test, I don't think we'd ever hit this logic block.

leerob:
Everything is working as expected 👍 
image

package.json
```diff
@@ -2,32 +2,89 @@
...
+  "type": "module",
+  "bin": {
+    "serve": "./build/main.js"
+  },
+  "files": [
+    "build/",
+    "package.json",
+    "readme.md",
+    "license.md",
+    ".npmrc"
```
styfle, line 22: I don't think `.npmrc` should be published

readme.md
```diff
@@ -0,0 +1,97 @@
+![](media/banner.png)
```
styfle, line 1: These images need to be absolute URLs or else they won't work on https://www.npmjs.com/package/serve or locally

readme.md
```diff
@@ -0,0 +1,97 @@
...
+
+<div align="center">
+  <a aria-label="Vercel logo" href="https://vercel.com">
+    <img src="https://img.shields.io/badge/made%20by-vercel-%23000000">
+  </a>
+  <br>
+  <a aria-label="Install Size" href="https://packagephobia.now.sh/result?p=serve">
```
styfle, line 8:
```suggestion
  <a aria-label="Install Size" href="https://packagephobia.com/result?p=serve">
```

readme.md
```diff
@@ -0,0 +1,97 @@
...
+
+![Listing UI](media/listing-ui.png)
+
+> Once it's time to push your site to production, we recommend using [Vercel](https://vercel.com).
+
+## Usage
+
+The quickest way to get started is to just run `npx serve` in your project's directory.
+
+If you prefer, you can also install the package globally using [Yarn](https://yarnpkg.com/en/) or [PNPM](https://pnpm.io/) (you'll need at least [Node LTS](https://github.com/nodejs/Release#release-schedule)):
```
styfle, line 31: I don't think we need to list every package manager. Lets just list `npm i -g serve` and the others can be assumed.

tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "@vercel/style-guide/typescript",
+  "compilerOptions": {
+    "target": "esnext",
+    "module": "esnext",
+    "moduleResolution": "node",
+    "noEmit": true,
```
styfle, line 7: I think we still need to emit JS

package.json
```diff
@@ -2,32 +2,89 @@
...
-  },
   "keywords": [
     "vercel",
     "serve",
     "micro",
     "http-server"
   ],
+  "repository": "vercel/serve",
   "license": "MIT",
+  "type": "module",
```
styfle, line 13: Are you sure we want to use ESM? Perhaps this needs an `engines` field so the consumer knows which versions of Node.js are supported.

tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "@vercel/style-guide/typescript",
+  "compilerOptions": {
+    "target": "esnext",
```
huozhi, line 4: target esnext requires quite new version of node and module esnext requires ESM. I feel we can keep the module as CJS output with es6 syntax, and support both ESM + CJS resolution in another PR

config/husky/pre-commit
mrmckeb: Colour libraries disable colours when `CI` is truthy, so this force-enables that.

Level 2 is 256 colour support.

source/utilities/server.ts
```diff
@@ -0,0 +1,126 @@
...
+    httpMode === 'https'
+      ? https.createServer(
+          {
+            key: await readFile(args['--ssl-key']),
+            cert: await readFile(args['--ssl-cert']),
+            passphrase: sslPass ? await readFile(sslPass, 'utf8') : '',
+          },
+          serverHandler, // eslint-disable-line @typescript-eslint/no-misused-promises
+        )
+      : http.createServer(serverHandler); // eslint-disable-line @typescript-eslint/no-misused-promises
```
mrmckeb, line 65: You could avoid this disable by not making `serverHandler` async, and adding a `catch` to the awaited `handler`.

source/utilities/config.ts
```diff
@@ -0,0 +1,126 @@
...
+    try {
+      switch (file) {
+        case 'now.json':
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+          parsedJson = (parsedJson as any).static;
+          break;
+        case 'package.json':
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+          parsedJson = (parsedJson as any).now.static;
+          break;
```
mrmckeb, line 82: Can we type these so we're not using `any`? We could also use optional chaining so that we don't have to rely on an error.

source/utilities/config.ts
```diff
@@ -0,0 +1,126 @@
...
+    // with the content in it.
+    const location = resolvePath(entry, file);
+
+    // Disabling the lint rule as we don't want to read all the files at once;
+    // if we can retrieve the configuration from the first file itself, we
+    // shouldn't waste time and resources fetching the other files too.
+    // eslint-disable-next-line no-await-in-loop
+    const [error, rawContents] = await resolve<string, NodeError>(
+      readFile(location, 'utf8'),
+    );
```
mrmckeb, line 45: Great explanation, thanks!

source/main.ts
```diff
@@ -0,0 +1,174 @@
...
+  process.exit(0);
+}
+if (args['--help']) {
+  logger.log(getHelpText());
+  process.exit(0);
+}
+
+// Default to listening on port 3000.
+// Disabling as this is not an unnecessary check.
+// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
mrmckeb, line 69: I think this error is correct, as type type of `args` (`Arguments`) extends `Options`, but none of those are optional.

Adding `Partial<Options>` will probably fix this for you.

source/utilities/http.ts
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Parse and return the endpoints from the given string.
+ *
+ * @param uriOrPort - The endpoint to listen on.
+ * @returns A list of parsed endpoints.
+ */
+export const parseEndpoint = (uriOrPort: ListenEndpoint): ParsedEndpoint => {
+  // If the endpoint is a port number, return it as is.
+  // @ts-expect-error `isNaN` accepts strings too.
+  if (!isNaN(uriOrPort)) return [uriOrPort];
```
mrmckeb, line 19: When you 
```suggestion
  if (!isNaN(Number(uriOrPort))) return [uriOrPort];
```

This then surfaces another issue which is why we ban `// @ts-ignore`. Can you please fix this in both places?

source/utilities/server.ts
```diff
@@ -0,0 +1,126 @@
...
+): Promise<ServerAddress> => {
+  // Define the request handler for the server.
+  const serverHandler = async (
+    request: IncomingMessage,
+    response: ServerResponse,
+  ): Promise<void> => {
+    if (args['--cors']) response.setHeader('Access-Control-Allow-Origin', '*');
+    // @ts-expect-error The `compression` library uses the Express Request type
+    // instead of the native HTTP IncomingMessage type.
+    if (!args['--no-compression']) await compress(request, response);
```
mrmckeb, line 46: This can be better solved with:
```suggestion
    type ExpressRequest = Parameters<typeof compress>[0];
    type ExpressResponse = Parameters<typeof compress>[1];
    if (!args['--no-compression']) await compress(request as ExpressRequest, response as ExpressResponse);
```

As disabling TS for the line will hide other errors.

source/utilities/server.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  // Listen for any error that occurs while serving, and throw an error
+  // if any errors are received.
+  server.on('error', (error) => {
+    throw new Error(`Failed to serve: ${error.stack}`);
+  });
+
+  // If the endpoint is a non-zero port, make sure it is not occupied.
+  // @ts-expect-error `isNaN` accepts strings too.
+  if (!isNaN(endpoint[0]) && endpoint[0] !== 0) {
```
mrmckeb, line 108: Again, here you can instead use `Number()` to parse the string.

source/utilities/server.ts
```diff
@@ -0,0 +1,126 @@
...
+    if (isClosed) return startServer([0], config, args, port);
+
+    // Otherwise continue on to starting the server.
+  }
+
+  // Finally, start the server.
+  return new Promise((resolve, _reject) => {
+    // @ts-expect-error I'm not sure why Typescript thinks this is invalid, the spread
+    // operator should pass it a `number` or a `string` or a `number`, `string`.
+    server.listen(...endpoint, () => resolve(getServerDetails()));
```
mrmckeb, line 124:
```suggestion
    server.listen(...endpoint as Parameters<typeof server.listen>, () => resolve(getServerDetails()));
```

mrmckeb, line 124: I don't think that TypeScript can determine which of the overloads you're trying to hit here, so that's why there's an error. This should fix it.

```suggestion
    server.listen(...endpoint as Parameters<typeof server.listen>, () => resolve(getServerDetails()));
```

source/utilities/cli.ts
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Parses the program's `process.argv` and returns the options and arguments.
+ *
+ * @returns The parsed options and arguments.
+ */
+export const parseArguments = (): Arguments =>
+  // @ts-expect-error The handler array for the `--listen` option does have only
+  // one element, I'm not sure why Typescript is throwing an error here.
+  parseArgv(options) as unknown as Arguments;
```
mrmckeb, line 123: It's because TS knows that an array is not always guaranteed.

You could remove the array or force the type:
```ts
  '--listen': parseEndpoint,
  '--listen': [parseEndpoint] as [typeof parseEndpoint],
```

This then surfaces another issue, which would be great to fix.
```
Argument of type '{ '--help': BooleanConstructor; '--version': BooleanConstructor; '--listen': (uriOrPort: ListenEndpoint) => ParsedEndpoint; ... 22 more ...; '-p': string; }' is not assignable to parameter of type 'Spec'.
  Property ''--listen'' is incompatible with index signature.
    Type '(uriOrPort: ListenEndpoint) => ParsedEndpoint' is not assignable to type 'string | Handler<any> | [Handler<any>]'.
      Type '(uriOrPort: ListenEndpoint) => ParsedEndpoint' is not assignable to type 'Handler<any>'.
        Types of parameters 'uriOrPort' and 'value' are incompatible.
```

mrmckeb (changes requested):
This looks great, a huge effort - thanks!

Let me know if you want help or want to pair on any of the comments, more than happy to help with any of it.

config/husky/pre-commit
leerob: Thank you!

source/main.ts
gamemaker1: Sure, we could add tests in a separate PR.

source/types.ts
gamemaker1: Done

source/utilities/config.ts
gamemaker1: It will only log a warning to the console if you use `package.json` or `now.json`, not throw an error.

I could remove support for these two locations, but we would then need to make a semver major release as it would be a breaking change.

package.json
gamemaker1: Oh that's cool - I didn't know that, will revert.

source/main.ts
leerob: https://github.com/vercel/serve/issues/709

source/utilities/config.ts
leerob: Yeah, agreed. Let's just leave them 👍

readme.md
gamemaker1: Using absolute URLs now, but they won't work until the PR is merged into the `main` branch.

readme.md
gamemaker1: Okay, will revert.

package.json
mrmckeb: No problem! That's what code reviews are all about ;)

readme.md
```diff
@@ -0,0 +1,96 @@
...
+  </a>
+  <br>
+  <a aria-label="Install Size" href="https://packagephobia.com/result?p=serve">
+    <img src="https://packagephobia.com/badge?p=serve">
+  </a>
+  <a aria-label="Stars" href="https://github.com/vercel/serve/stargazers">
+    <img src="https://img.shields.io/github/stars/vercel/serve">
+  </a>
+  <a aria-label="Build Status" href="https://github.com/vercel/serve/actions/workflows/ci.yaml">
+    <img src="https://github.com/gamemaker1/serve/actions/workflows/ci.yaml/badge.svg">
```
mrmckeb, line 15: This should be `vercel` and not `gamemaker1` here, sorry I missed this before!

tsconfig.json
gamemaker1: I'm using `tsc` only to check types, `tsup` does the actual compilation of ts => js, so there is no need for `tsc` to emit js code.

package.json
gamemaker1: Yes, we will need to use ESM as a few modules like `chalk` are now ESM only. Also, it allows us to use top level `await`.

gamemaker1: Maybe we could target Node 14 instead of Node 18, so we change `target` in tsconfig.json to es2019.

tsconfig.json
gamemaker1: What do you think about this?

https://github.com/vercel/serve/pull/706#discussion_r917546684

I don't think we can output CJS as some of the modules (e.g., `chalk`) we use are pure ESM.

gamemaker1:
I have to step away for now, but once I am bacj I will address the `ts-expect-error` and `eslint-disable` comments as @mrmckeb pointed out. Thanks for the explanation and the hints @mrmckeb!

config/husky/pre-commit
gamemaker1: It seems git (or pnpm?) disables color output when running the hooks, so this force-enables it.

This happens locally, not in CI.

package.json
```diff
@@ -2,32 +2,89 @@
...
+      "./node_modules/@vercel/style-guide/eslint/typescript.js"
+    ],
+    "parserOptions": {
+      "project": "tsconfig.json",
+      "warnOnUnsupportedTypeScriptVersion": false
+    },
+    "rules": {
+      "@typescript-eslint/restrict-template-expressions": 0,
+      "@typescript-eslint/no-explicit-any": 0
+    }
```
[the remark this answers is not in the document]
gamemaker1: Removed them, and fixed the lint errors.

source/utilities/server.ts
gamemaker1: Ahh okay, will do!

tsconfig.json
huozhi: It brings quite a lot changes though IMO, you don't need to bump all the deps together. If we mark this repo as `module: "commonjs"` with no deps bumping, it can still be imported by ESM package

package.json
huozhi: For `chalk`, do we have to upgrade it to the ESM only version? Can stay with the current used version so that it doesn't cause break for old node versions?

gamemaker1: Old Node versions that don't support ESM are no longer even in maintainance mode (i.e., Node 12 and before). If someone wants to use `serve` with those versions, they could use older versions of `serve`. With ESM, we can target Node 14 onwards and use quite a few more features and libraries.

It's not just `chalk` - `clipboardy` and `is-port-reachable` are also ES modules. We would need to either downgrade, replace or write those modules ourselves if we want to use CJS - introducing more potential bugs or security issues.

source/utilities/config.ts
gamemaker1: Done!

source/utilities/server.ts
gamemaker1: I wrapped the contents of the `serverHandler` function in an async function instead, as the `compress` call was also async.

source/main.ts
gamemaker1: Good catch, using `Partial<Options>` did work. Thanks!

source/utilities/server.ts
gamemaker1: Fixed it by manually checking which signature to use.

source/utilities/http.ts
gamemaker1: Fixed it ;]

source/utilities/cli.ts
gamemaker1: The function signature of `parseEndpoint` wasn't compatible with the `Handler` signature, so I fixed that.

package.json
gamemaker1: Changed the `target` field in `tsconfig.json` to ES2020 - which means the compiled code supports Node 14 onwards.

package.json
```diff
@@ -2,32 +2,89 @@
...
+  },
+  "prettier": "@vercel/style-guide/prettier",
+  "eslintConfig": {
+    "extends": [
+      "./node_modules/@vercel/style-guide/eslint/node.js",
+      "./node_modules/@vercel/style-guide/eslint/typescript.js"
+    ],
+    "parserOptions": {
+      "project": "tsconfig.json",
+      "warnOnUnsupportedTypeScriptVersion": false
```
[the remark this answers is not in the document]
gamemaker1: I'll switch typescript to v4.6.4 then.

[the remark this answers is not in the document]
gamemaker1: Done!

readme.md
gamemaker1: Done :]

gamemaker1:
Removed all the `ts-expect-error` and `eslint-disable` comments that @mrmckeb pointed out. Thanks for pointing it out and helping me remove them!

package.json
leerob: I'm okay with Node 14+ support 👍

tsconfig.json
leerob: The plan is to publish this as a breaking change.

gamemaker1: That sounds like a good idea.

leerob (approved):
All of my comments have been addressed. To reiterate, we will be releasing this as a major version and can write a nice summary in the release notes to let folks know what changed.

Let's get two other +1s and we can ship it.

gamemaker1:
That's great!

Once this PR is merged, I would love to help out with adding the tests and some features from forks like https://github.com/warren-bank/node-serve; looking forward to collaborating with all of you on it :]

tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "@vercel/style-guide/typescript",
+  "compilerOptions": {
+    "lib": ["es2020"],
+    "target": "es2020",
+    "module": "esnext",
+    "moduleResolution": "node",
+    "noEmit": true,
+    "resolveJsonModule": true,
+    "esModuleInterop": true
```
mrmckeb, line 10: I didn't notice this last time sorry - this is already on in the shared style-guide.

That being said, no real issue with it being redefined here... so only if you're pushing any other changes.

mrmckeb (approved):
This is great. I'm not an expert on this package, but the changes you've made look logical and you've done a great job of typing and adhering to our style guide - thank you!

tsconfig.json
gamemaker1: I hadn't noticed this before, thanks for pointing it out! Removed it in 083a43c.

leerob:
Thank you @gamemaker1 🙏 Would love to send you some Vercel swag (shoot me an email `lee@vercel.com`).

leerob:
Confirmed things are working as expected, publishing stable – will keep an eye on issues 😄

leerob:
So far I've only seen some small bug: https://github.com/vercel/serve/issues/711

gamemaker1:
> Would love to send you some Vercel swag (shoot me an email).

Wow, thank you! I will send you an email today :]

Issue #508, opened 2019-02-11
Keep running serve when terminal window closed


sinclas:
Hi,

How can I keep serve running when the terminal window closes.  should I use a systemd script as a service?  If so, what would the path to serve be?

Thanks!
Steve

r0mflip:
If you are trying to run as a "service" then do use the systemd scripts and use absolute path to the folder. Call the service from within the terminal.

pavanmehta91:
You can use some utility like `screen` for linux based systems.
https://linuxize.com/post/how-to-use-linux-screen/

PR #708, opened 2022-07-09
Add support for PFX or PKCS12 encoded certificates

schemburkar:
_This PR adds additional work based on the excellent work in #706 for Typescript._

Added support for PFX or PKCS12 encoded certificates by auto detecting format using options passed for SSL. No additional parameter needed to detect. This continues the work in #508 that was in older JS format

For using PFX certificates use following parameters
`--ssl-cert` This is the path to the PFX certificate
`--ssl-pass` (_Optional_) This is the path to the password for PFX certificate.

Examples for PFX

```bash
serve  --ssl-cert ./cert.pfx --ssl-pass ./password.txt 
```

```bash
serve  --ssl-cert ./cert.pfx 
```

For PEM (no change)

```bash
serve  --ssl-cert ./cert.crt --ssl-key ./certkey.key --ssl-pass ./password.txt
```

**This PR should only be merged after #706.**
CC: @leerob @gamemaker1


leerob:
Linked PR is now merged 🎉 Want to merge in those changes and let me know when this is ready for review?

source/utilities/server.ts
```diff
@@ -101,6 +111,7 @@ export const startServer = async (
       else address = details.address;
       const ip = getNetworkAddress();
 
+      const httpMode = useSsl ? 'https' : 'http';
```
gamemaker1, line 114: Could you rename this to `protocol`, since the variable is being used as the protocol of the server URL.

schemburkar: Thanks for the feedback, this is done

leerob (approved):
Would love a test for this – maybe we can add that in follow-up building off https://github.com/vercel/serve/pull/712.

leerob:
Testing has now been added!

schemburkar:
@leerob I have updated the cli test snapshot with updated text. Can I start another PR for adding tests specific for PFX certs? This will help get this PR merged while I work on tests. 

Open to suggestions

PR #727, opened 2022-08-29
Add no-cache flag

querielo:
The PR allows us to disable browser cache. It can be suitable if we often change served content.

### DIFF
--- a/source/utilities/server.ts
+++ b/source/utilities/server.ts
@@ -63,6 +63,8 @@ export const startServer = async (
 
       if (args['--cors'])
         response.setHeader('Access-Control-Allow-Origin', '*');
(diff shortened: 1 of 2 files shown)


davidmroth (approved):
Changes look good!

source/utilities/server.ts
```diff
@@ -63,6 +63,8 @@ export const startServer = async (
 
       if (args['--cors'])
         response.setHeader('Access-Control-Allow-Origin', '*');
+      if (args['--no-cache'])
```
leerob, line 66: Could we add a test please?

LinusU:
This would be much appreciated 🙏

PR #730, opened 2022-09-14
fix: mute error by default

### LINKED ISSUE #677: Regression, Red error is thrown in Js only environments
This is also a problem on CodeSandbox, Runkit and Glitch... where unless specially configured, an attention consuming red Error is thrown. We can only use Javascript in codesandbox... apt does not work. Before the clipboard feature and the PR's to fix it... this was not a problem.

Screenshot from 2021-10-19 09-49-23

## Potential solutions

- Mute the error by default. Unless debug/verbose mode is on.
- Upgrade to clipboardy 3.0? it includes a binary fallback for linux.

## Related
 - #533 (Should be closed)
 - #623 original WSL issue
 - #612 (PR to fix the WSL issue)
 - https://github.com/sindresorhus/clipboardy/issues/63

ZanMinKian:
Closed #677

### DIFF
--- a/source/main.ts
+++ b/source/main.ts
@@ -133,9 +133,10 @@ for (const endpoint of args['--listen']) {
       await clipboard.write(local);
       message += `\n\n${chalk.grey('Copied local address to clipboard!')}`;
     } catch (error: unknown) {
-      logger.error(
-        `Cannot copy server address to clipboard: ${(error as Error).message}.`,
-      );
+      if (args['--debug'])
+        logger.error(
+          `Cannot copy server address to clipboard: ${(error as Error).message}.`,
+        );
     }
   }
 


source/main.ts
```diff
@@ -133,9 +133,10 @@ for (const endpoint of args['--listen']) {
       await clipboard.write(local);
       message += `\n\n${chalk.grey('Copied local address to clipboard!')}`;
     } catch (error: unknown) {
-      logger.error(
-        `Cannot copy server address to clipboard: ${(error as Error).message}.`,
-      );
+      if (args['--debug'])
```
leerob, line 136: Is there any way to tell this is being used in an environment where this isn't wanted?

Also, could we add a test please?

Deccal383838 (approved):

PR #743, opened 2022-11-08
infra: move c8 (coverage) to devDeps

AviVahl:
- used only in tests, but caused downstream to install it as well, raising the number of deps.
- also bumped version to 7.12.0 (latest)


AviVahl:
@AndyBitz thanks for the security-fix release. I've noticed this little (dep size) regression, so here's a quick PR to address it.

AndyBitz (approved):
Amazing! Thank you

AviVahl:
Awesome. Thanks again 🙏

PR #745, opened 2022-11-09
Print QR code at serve startup

citizenll:
>usage
```bash
> serve -q
```
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "serve",
-  "version": "14.1.2",
+  "version": "14.2.0",
   "description": "Static file serving and directory listing",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -44,6 +45,7 @@
     "clipboardy": "3.0.0",
     "compression": "1.7.4",
     "is-port-reachable": "4.0.0",
+    "qrcode-terminal": "0.12.0",
```
leerob, line 48: How much extra size does this add to the package?

Could we add a test please?

Deccal383838 (approved):

marcofugaro:
oh this would be useful!

PR #753, opened 2023-01-08
Update CORS headers to support PNA spec

k-yle:
Serve is currently broken in Microsoft Edge (strict mode), since Edge is trialling the rollout of the [Private Network Access (PNA)](https://developer.chrome.com/blog/private-network-access-preflight/) spec. Chrome is planning on rolling this out soon, so it's only a matter of time before this affects more users. 

These new headers allow non-localhost websites to make fetch requests to http://localhost

### DIFF
--- a/source/utilities/server.ts
+++ b/source/utilities/server.ts
@@ -61,8 +61,11 @@ export const startServer = async (
           chalk.cyan(requestUrl),
         );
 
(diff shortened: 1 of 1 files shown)


leerob (approved):
Thanks!

leerob:
Would be awesome to add a test!

PR #763, opened 2023-04-16
Ignore virtual network switches (fixes issue #761)

### LINKED ISSUE #761: Local Network ip prefers hyper-v virtual switch over actual network
### Description

WindowsTerminal_pJvT4H1m9E
Code_llGEHcQLCJ

In the attachment is shown it has two switches, the hyper-v default switch and a custom wsl switch, in the current use case it's the vEthernet default switch we're talking about.
Right now, when selecting local network, instead of using the proper ip address, it wants to make use of the hyper-v switch, which shouldn't be the case.

To reproduce, you should install hyper-v on your windows (11) machine. After setting up a default switch, which should happen when you either create your first virtual machine or start the hyper-v service, serve any node application.

Right now I used a temporary solution by manually disabling the eth interface through an administrator-ran powershell command:
`get-netadapter -IncludeHidden|where-object {$_.interfacedescription -like "*hyper-v*"}|Disable-NetAdapter`

I suggest that as a fix to either add an argument or setting to explicitly enable including hyper-v switches, or just filtering them out when listing the network interfaces at all.

### Library version

14.2.0

### Node version

19.4.0

Ivanmatthew:
Resolves #761

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "serve",
-  "version": "14.2.0",
+  "version": "14.2.1",
   "description": "Static file serving and directory listing",
   "keywords": [
     "vercel",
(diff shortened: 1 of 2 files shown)


Ivanmatthew:
Update?

Issue #792, opened 2023-12-22
add --open to the CLI


Pomax:
### Description

It'd be nice if there was a `--open` flag that just ran [open](https://www.npmjs.com/package/open) with the server URL, so folks don't need to either copy-paste URL strings, or set up parallel scripts that have to guess at the URL that needs to be opened.

### Why

Because it's one less "tiny bother that adds up over the course of a day's worth of testing" thing folks need to do in order to just focus on their work.

### Alternatives

everyone who needs it can manually install `open`, add it to their npm script, and then run `serve` and `open` in parallel. However, the "in parallel" part is not exactly trivial (with npm-run-all being deprecated) and you'd have to "guess" the server URL. It's _probably_ localhost:3000 but that's configurable so it might not be. Also, looking at the dependency list for `serve`, adding `open` to that would effectively have zero impact on `serve`'s footprint while making dev life just that bit easier.

JakedUp:
Agreed on this. Would be awesome to have this.

PR #798, opened 2024-01-30
feat: open in browser

gnekich:
This PR implements #792, introduces new feature to open URL in browser with `--open` flag and also introduces shortcuts the same way [vitejs](https://vitejs.dev/) does it.

Example01


Dex19:
This is a very practical feature, kudos!

PR #814, opened 2024-10-22
Add Access-Control-Expose-Headers in --cors mode

jtbandes:
By default, browsers only allow clients to read [CORS-safelisted response headers](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_response_header). However, sometimes it is [useful to read](https://github.com/cruise-automation/webviz/blob/d13afdacaec0b48f983adcaca55b84c36c42a5f2/packages/webviz-core/src/dataProviders/BrowserHttpReader.js#L39-L41) other headers that are not safelisted by default, such as `Accept-Ranges`. The [`Access-Control-Expose-Headers`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers) header allows controlling which headers are exposed to the client.

This change exposes all headers (`*`) when `--cors` is used.

It could be useful to allow more fine-grained configuration of exposed headers, however, it seems that the existing `--cors` mode is a very coarse toggle so I just used `*` for consistency.

I didn't see a relevant place to update docs – is this tool missing documentation for the CLI flags?


jtbandes:
Friendly ping @AndyBitz @leo  -- any interest in this patch?

jtbandes:
Hi there, just checking in again if there's any interest here?

PR #818, opened 2025-01-01
Update license.md, fix copyright license year

JasonnnW3000:
Yearly copyright bump. Padding my changed LOC! 😛

### DIFF
--- a/license.md
+++ b/license.md
@@ -1,6 +1,6 @@
 # The MIT License (MIT)
 
-Copyright (c) 2023 Vercel, Inc.
+Copyright (c) 2025 Vercel, Inc.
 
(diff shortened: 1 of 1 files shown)


MikeMcC399:
This PR is unnecessary - see https://nexb.com/do-you-really-need-to-update-the-copyright-each-new-year/

I believe it is also incorrect, as the Copyright should include the first year of the work, so if anything it should be at least a range covering what it previously covered:

> Copyright (c) 2023-2025 Vercel, Inc.

Also, since there have been no changes yet to the repo contents, there is not yet anything to copyright for the year 2025, which makes it additionally premature.

Probably this PR can be closed.

PR #848, opened 2026-03-12
fix: replace clipboardy with tinyclip

gameroman:


florian-lefebvre:
For context, tinyclip is [really lighter](https://npmx.dev/compare?packages=tinyclip,clipboardy) than clipboardy. This PR was made as part of the [e18e initiative](https://e18e.dev/).

florian-lefebvre:
Node engines would need to be bumped most likely

styfle (approved):
Thanks!

florian-lefebvre:
Looks like this needs a changeset so it can be released

PR #851, opened 2026-04-14
feat: add QR code for server URLs in terminal

showayebDev:
### Description
This PR adds a QR code display for both Local and Network URLs when the server starts. 

### Why?
Manually typing long local IP addresses on mobile devices for testing is tedious. Scanning a QR code from the terminal makes cross-device testing significantly faster and more convenient.

### Changes
- Added `qrcode-terminal` as a dependency.
- Created `source/utilities/qrcode.ts` utility.
- Updated `source/main.ts` to render QR codes inside the startup box.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,10649 @@
+{
+  "name": "serve",
+  "version": "14.2.6",
+  "lockfileVersion": 3,
+  "requires": true,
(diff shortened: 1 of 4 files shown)


kkhcheng:
why is this not merge?  this would be so useful!

